PR #13, opened 2016-04-26
Requests

huebnerob:
https://vimean.atlassian.net/browse/TVOS-43

This PR adds some more high-level template requests to our library.  From here, we can probably just add any more as needed.


huebnerob:
@alfiehanssen @mikew-personal @jasonhawkins Can I get a review from somebody? 🙏

jasonhawkins:
I can review!

jasonhawkins:
I had to update the bridging header to get this to build. It looks like this was a problem before too, but can you change the imports in `VimeoNetworking-Bridging-Header.h` to: 
```ObjC
#import <VimeoNetworking/VIMUser.h>
#import <VimeoNetworking/VIMAccountNew.h>
#import <VimeoNetworking/VIMVideo.h>
```

jasonhawkins:
What's the best way for me to test this? The tvOS project?

huebnerob:
Thanks @jasonhawkins! 

You could mess with making some requests somewhere in the sample project and ensure they return expected results.

jasonhawkins:
I'm getting 400's when I run the sample project just as it is. Is that how it's been/supposed to be?

huebnerob:
It might be, depending on which authentication method is currently commented/uncommented.  Some of the test requests are trying the `/me` endpoint which won't work if we're using client credentials grant.  I'd recommend replacing everything in the app delegate `testEndpoints()` function with whatever you want to try out.

VimeoNetworking/VimeoNetworking/Request+Category.swift
```diff
@@ -0,0 +1,27 @@
...
+import Foundation
+
+public typealias CategoryRequest = Request<VIMCategory>
+public typealias CategoryListRequest = Request<[VIMCategory]>
+
+public extension Request
+{
+    private static var CategoriesPath: String { return "/categories" }
+    
+    public static func getCategoryRequest(categoryURI categoryURI: String) -> Request
```
jasonhawkins: How do you feel about adding documentation for these public methods?

VimeoNetworking/VimeoNetworking/Request+Channel.swift
```diff
@@ -0,0 +1,33 @@
...
+    
+    public static func queryChannels(query query: String, refinements: VimeoClient.RequestParameters? = nil) -> Request
+    {
+        var parameters = refinements ?? [:]
+        
+        parameters[self.QueryKey] = query
+        
+        return Request(path: self.ChannelsPath, parameters: parameters)
+    }
+}
```
jasonhawkins: Git wants new lines at the end of all of these.

VimeoNetworking/VimeoNetworking/Request+User.swift
```diff
@@ -35,4 +47,46 @@ public extension Request
...
+    
+    public static func getUserFollowersRequest(userURI userURI: String) -> Request
+    {
+        return Request(path: String(format: self.FollowersPathFormat, userURI))
+    }
+    
+    // MARK: - Search
+    
+    
+
```
jasonhawkins: Extra whitespace.

VimeoNetworking/VimeoNetworking/Request+Category.swift
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return Request(path: categoryURI)
+    }
+    
+    public static func getCategoriesRequest() -> Request
```
jasonhawkins: Do these requests need to be authenticated for them not to fail? Are there any that don't require authentication?

jasonhawkins:
Looks like you'll need to pull in the latest from master here.

VimeoNetworking/VimeoNetworking/Request+Channel.swift
alfiehanssen: Is that what that means? Why does it want that? 😨

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -10,20 +10,36 @@ import Foundation
...
 public enum CacheFetchPolicy
 {
-    static let DefaultPolicy: CacheFetchPolicy = .CacheThenNetwork
-    
     case CacheOnly
     case CacheThenNetwork
     case NetworkOnly
     case TryNetworkThenCache
+    
+    static func defaultPolicyForMethod(method: VimeoClient.Method) -> CacheFetchPolicy
```
alfiehanssen: oooooh 👍

VimeoNetworking/VimeoNetworking/Request+Category.swift
huebnerob: Everything needs an access token (client credentials, minimum) but, no, these don't need you to be logged in as a user

PR #20, opened 2016-05-05
Swift header fix

huebnerob:
This branch fixes an issue where an erroneous `#import <VimeoNetworking/VimeoNetworking.h>` line was added to the autogenerated `VimeoNetworking-Swift.h` file in any projects which integrated VimeoNetworking using CocoaPods.  This error was introduced when we added the `ExceptionCatcher` logic to the library, and the proposed changes clarify some of the architecture around that bit of code which seems to confuse the compiler a bit less.

The logic behind why this issue was caused by this change is still unclear, but it has something to do with the confluence of the following factors:
- `ExceptionCatcher`, the class, was declared in objective c, and contained one (underscored) method which took a void block to execute and returned a nullable error.
- a Swift extension to `ExceptionCatcher` was declared so we could use the `throws` error-handling pattern.  This had a (non-underscored) `doUnsafe` function which was intended for swift use only.
- Despite only ever actually wanting to use the extension function in swift, the compiler still insisted on bridging it back to ObjC for compatibility, and the generated method signature was something along the lines of: `- (BOOL)doUnsafeAndReturnError:(NSError * __nullable * __nonnull)error :(Void(^)void)voidBlock;`.

Since we don't actually need that (albeit clever) translation of the swift throwable function extension back to objective-c, this branch inhibits the compiler from doing just that, and in turn squashes the weird module header import problem mentioned above.

### DIFF
--- a/VimeoNetworking.podspec
+++ b/VimeoNetworking.podspec
@@ -20,7 +20,7 @@ Pod::Spec.new do |s|
 
   s.source = { :git => "https://github.com/vimeo/VimeoNetworking", :tag => s.version.to_s }
 
-  s.source_files  = "VimeoNetworking/VimeoNetworking"
+  s.source_files  = "VimeoNetworking/VimeoNetworking/*.{h,m,swift,cer}"
   s.exclude_files = "VimeoNetworking/VimeoNetworking/VimeoNetworking.h"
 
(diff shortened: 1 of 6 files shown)


alfiehanssen:
👍 DWR 👍

PR #21, opened 2016-05-11
CAM-265: Modified loadSavedAccount logic

alfiehanssen:
#### Ticket
[CAM-265](https://vimean.atlassian.net/browse/CAM-265)

#### Ticket Summary
I'm building out a revised user system for the Cameo-iOS app. As part of this, we're leveraging the VimeoNetworking accounts system. Because of differences with how the Cameo-iOS app treats users, we don't want to call `loadSavedAccount`, but prefer to be able to be more explicit by calling a method like `loadUserAccount` or `loadClientCredentialsAccount`. 

#### Implementation Summary
I made the above changes. Introducing these two new methods. Modifying existing logic to leverage these two new methods. And marked `loadSavedAccount` as deprecated because it sounds like we're moving to a more specific API, away from the general-ness of `loadSavedAccount` (because it attempts to load a user account, and if it can't it then attempts to load a client credentials account, this feels opaque).

#### How to Test
Make sure accounts are being loaded as expected (in the demo project).

### DIFF
--- a/VimeoNetworking/VimeoNetworking/AuthenticationController.swift
+++ b/VimeoNetworking/VimeoNetworking/AuthenticationController.swift
@@ -48,35 +48,58 @@ final public class AuthenticationController
         self.authenticatorClient = VimeoClient(appConfiguration: client.configuration)
     }
     
-    // MARK: - Saved Accounts
+    // MARK: - Public Saved Accounts
     
+    public func loadClientCredentialsAccount() throws -> VIMAccount?
+    {
+        return try self.loadAccount(.ClientCredentials)
+    }
+
+    public func loadUserAccount() throws -> VIMAccount?
+    {
+        return try self.loadAccount(.User)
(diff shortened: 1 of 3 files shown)


huebnerob:
looks good! 👍

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -48,35 +48,58 @@ final public class AuthenticationController
...
+    {
+        return try self.loadAccount(.ClientCredentials)
+    }
+
+    public func loadUserAccount() throws -> VIMAccount?
+    {
+        return try self.loadAccount(.User)
+    }
+    
+    @available(*, deprecated, message="Use loadUserAccount or loadClientCredentialsAccount instead.")
```
huebnerob: 🔥

PR #22, opened 2016-05-11
TVOS-150: Inline Documentation

huebnerob:
# Ticket(s)

https://vimean.atlassian.net/browse/TVOS-150

# Ticket Summary

Document all public types, functions, and properties 

# Implementation Summary

- Wrote instructions for humans
- Some very light code reorganization where necessary

# How to test

Prepare the reading monocle :tophat:


huebnerob:
Hitting you hard with all these PRs @jasonhawkins but you're the docs guy!  Check it out too when you have a sec @alfiehanssen @ghking @nicolelehrer @mikew-personal

jasonhawkins:
🍕I AM SO EXCITED 🍕

VimeoNetworking/VimeoNetworking/AccountStore.swift
```diff
@@ -22,13 +27,28 @@ final internal class AccountStore
...
+     */
     init(configuration: AppConfiguration)
     {
         self.dataStore = KeychainStore(service: configuration.keychainService, accessGroup: configuration.keychainAccessGroup)
     }
     
     // MARK: -
     
+    /**
+     Securely save an account
```
jasonhawkins: It might be helpful to make it clear why this is secure.

VimeoNetworking/VimeoNetworking/AccountStore.swift
```diff
@@ -39,6 +59,15 @@ final internal class AccountStore
         try self.dataStore.setData(data, forKey: type.rawValue)
     }
     
+    /**
+     Securely load an account
```
jasonhawkins: Here too.

VimeoNetworking/VimeoNetworking/AccountStore.swift
```diff
@@ -65,6 +94,13 @@ final internal class AccountStore
         return account
     }
     
+    /**
+     Removes a saved account
```
jasonhawkins: And is removing an account _not_ secure?

VimeoNetworking/VimeoNetworking/AppConfiguration.swift
```diff
@@ -8,11 +8,17 @@
 
 import Foundation
 
-public let KeychainServiceVimeo = "Vimeo"
-public let KeychainServiceCameo = "Cameo"
-
+/**
+ *  Stores all static information relevant to a client application
```
jasonhawkins: 🤘🏻

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -8,8 +8,12 @@
 
 import Foundation
 
+/// `AuthenticationController` is used to authenticate a `VimeoClient` instance, either by loading an account stored in the system keychain, or by interacting with the Vimeo API to authenticate a new account.  The two publicly accessible authentication methods are client credentials grant and code grant.  Client credentials grant is a public authentication method with no logged in user, which allows your application to access anything a logged out user could see on Vimeo (public content).  Code grant is a way of logging in with a user account, which can give your application access to that user's private content and permission to interact with Vimeo on their behalf.  This is achieved by first opening a generated URL in Safari which presents a user log in page.  When the user authenticates successfully, control is returned to your app through a redirect link, and you then make a final request to retrieve the authenticated account.
```
jasonhawkins: What if we break this up a little bit like:
```Swift
/**
 `AuthenticationController` is used to authenticate a `VimeoClient` instance, either by loading an account stored in the system keychain, or by interacting with the Vimeo API to authenticate a new account.  The two publicly accessible authentication methods are client credentials grant and code grant.
 
 Client credentials grant is a public authentication method with no logged in user, which allows your application to access anything a logged out user could see on Vimeo (public content).
 
 Code grant is a way of logging in with a user account, which can give your application access to that user's private content and permission to interact with Vimeo on their behalf.  This is achieved by first opening a generated URL in Safari which presents a user log in page.  When the user authenticates successfully, control is returned to your app through a redirect link, and you then make a final request to retrieve the authenticated account.
 
 - Note: This class contains implementation details of private Vimeo-internal authentication methods in addition to the public authentication methods mentioned above.  However, only client credentials grant and code grant authentication are supported for 3rd-party applications interacting with Vimeo.  None of the other authentication endpoints are enabled for non-official applications, nor will they ever be, so please don't attempt to use them.  Perhaps you're thinking, "Why are they here at all, then?"  Well, Vimeo's official applications use this very same library to interact with the Vimeo API, and this, more than anything, is what keeps `VimeoNetworking` healthy and well-maintained :).
*/
```

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -160,38 +200,75 @@ final public class AuthenticationController
...
     // MARK: - Private Authentication
     
+    /**
+     **(PRIVATE: Vimeo Use Only)**
+     Log in with an email and password
+     
+     - parameter email:      a user's email
+     - parameter password:   a user's password
+     - parameter completion: handler for authentication success or failure
+     */
```
jasonhawkins: I think it might be worth adding something like `- Warning: This method will not work for third party applications.` to all of these private Vimeo methods. Or at least some way to indicate that it's not just that we don't want people to use it. 

It's that they _can't_.

VimeoNetworking/VimeoNetworking/ErrorCode.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 
+/// `VimeoErrorCode` contains all api error codes that are currently recognized by our client applications
```
jasonhawkins: I 💟 everything about this. Documented error codes are 🎆.

VimeoNetworking/VimeoNetworking/VimeoResponseSerializer.swift
```diff
@@ -84,6 +106,13 @@ final public class VimeoResponseSerializer: AFJSONResponseSerializer
         try self.checkStatusCodeValidity(response: response)
     }
 
+    /**
+     Check a that a download task response has a valid status code
```
jasonhawkins: `Check that a download task…`

jasonhawkins:
It looks like you've got a conflict with `master` but I done reviewing!

VimeoNetworking/VimeoNetworking/AccountStore.swift
```diff
@@ -65,6 +94,13 @@ final internal class AccountStore
...
+     
+     - parameter type: type of account to remove
+     
+     - throws: an error if the data could not be removed
```
alfiehanssen: What happens if there is no data to remove?

VimeoNetworking/VimeoNetworking/AppConfiguration.swift
```diff
@@ -8,11 +8,17 @@
...
+ */
 public struct AppConfiguration
 {
+        /// Keychain identifier for the Vimeo application
+    public static let KeychainServiceVimeo = "Vimeo"
```
alfiehanssen: Should we add tvOS?

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
alfiehanssen: Can we use the `@available` annotation similar to how we add deprecation warnings?

VimeoNetworking/VimeoNetworking/NSError+Extensions.swift
```diff
@@ -86,22 +101,52 @@ public extension NSError
...
     }
     
+    /**
+     Adds information to a preexisting error
+     
+     - parameter domain:   New domain for the error
+     - parameter code:     the new error code for the error
+     - parameter userInfo: the user info dictionary to append to the existing user info
+     
+     - returns: The error with additional information in the user info dictionary
```
alfiehanssen: Worth mentioning that this and other methods in this class return a new error object rather than augmenting the existing one?

VimeoNetworking/VimeoNetworking/AccountStore.swift
huebnerob: Good point, that's not an error, augmented the docs to reflect this

VimeoNetworking/VimeoNetworking/AppConfiguration.swift
huebnerob: Isn't it technically the same application by bundle identifier? I think they can both just use Vimeo

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
huebnerob: I'll augment the docs to be more clear about it not working for third parties.  I'll stop short of adding an actual warning, though, because that'd clutter our own projects :/

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
huebnerob: 🎯

VimeoNetworking/VimeoNetworking/NSError+Extensions.swift
huebnerob: 👍

VimeoNetworking/VimeoNetworking/AppConfiguration.swift
alfiehanssen: 👍

PR #30, opened 2016-05-26
cam-xxx-disable-request-cache

mikew-personal:
#### Ticket
N/A

#### Ticket Summary
The VimeoSessionManager is using the default NSURLSessionConfiguration which has a request cache policy of .UseProtocolCachePolicy.  This is causing certain API calls (ex. /me/videos) to return back cached data instead of fetching new results from the server, which is resulting in the wrong data being returned when logging in with different user accounts. 

#### Implementation Summary
I created some convenience methods in an extension of the NSURLSessionConfiguration class.  These methods allow you to create a new default session configuration by passing in the desired caching policy.  I modified the code in VimeoSessionManager+Constructors to use the new convenience methods to create default session configurations that have cache policies set to .ReloadIgnoringCacheData, which will disable the URL cache.

#### How to Test
1) Log in with a user account and request the list of videos from the /me/videos endpoint
2) perform the request again, this time cancelling the request
3) log out and log in as a different user
4) perform the request again, verify that the response was not retrieved from the NSURL cache.

### DIFF
--- a/VimeoNetworking/VimeoNetworking.xcodeproj/project.pbxproj
+++ b/VimeoNetworking/VimeoNetworking.xcodeproj/project.pbxproj
@@ -7,11 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


huebnerob:
Love it 👍

PR #31, opened 2016-05-27
[WIP] Fix Cameo build issues

huebnerob:
With these changes, we no longer have a compiler segmentation fault when we uncomment the erroneous VimeoNetworking.h import in the swift generated header, but we're still getting that header.  Working on fixing the latter problem.


alfiehanssen:
@huebnerob Is this ready for review? Can you assign someone? Is it the same material that a user pull request recently (PR is still open)?

huebnerob:
@alfiehanssen I'm forgetting where we landed with these, are we still seeing these Cameo build issues?

alfiehanssen:
I don’t think there’s an issue with Cameo any more

PR #37, opened 2016-06-15
Fix #35: VimeoClient currentAccount accessible outside module

### LINKED ISSUE #35: Single user application, lightweight use
#### Issue Summary
It should be possible to use the token generated for an API app (https://developer.vimeo.com/apps) to authenticate api calls issued by the VimeoClient. On the VIMNetworking repo this technique seemed to be referred as [Lightweight Use](https://github.com/vimeo/VIMNetworking#lightweight-use).

#### Reproduction Steps
- Go to https://developer.vimeo.com/apps
- Create an app or use an existing one
- Generate an Access Token
- On your application try to use the generated access token as a way to authenticate api calls.

#### Expected Behavior
VimeoClient class provide a way to set the generated access token. Following api calls use this token.

#### Actual Behavior
As of today we can only authenticate using the two approaches mentioned in the README.md, client credentials or code grant. 

One quick (but dirty?) fix could be to modify the access level of the currentAccount property in the VimeoClient class. With this modification users of the library would be able to write the following code

```
let vimeoClient = VimeoClient(appConfiguration: appConfiguration)
vimeoClient.currentAccount = VIMAccount(keyValueDictionary: ["accessToken" : "generated acess token", "tokenType" : "Bearer", "scope" : "scope associated with access token"])
let videoRequest = Request<VIMVideo>(path: "/videos/video_id")
``` 

and hence bypass the authentication controller. If this is something that you would consider let me know and I will create a PR for this fix.

nicolasabric:


huebnerob:
@nicolasabric I'd actually prefer to keep the `currentAccount` property immutable from the perspective of other modules, if possible.  An `AuthenticationController` should be used to authenticate a `VimeoClient`.  Could you elaborate on why this is required in your setup?

nicolasabric:
@huebnerob yes sure. We are currently working on tvOS application where people will be able to watch videos and then get some info about the video's location, prices to go there, etc. Technically we are doing the following:
- We have a Vimeo pro account where we uploaded some videos,
- We have created an API app connected to our pro account and retrieved a Bearer from there,
- Our tvOS application retrieve the videos through the /me/videos api endpoint. Ideally using the bearer in previous step.
- Users play videos within our tvOS app. Therefore files property in API's response is mandatory.

Obviously we don't want the users of our app to have a Vimeo account (PRO or otherwise) instead we want our application to automatically login as our Vimeo pro account (hence usage of the API app + Bearer). 

It's my current understanding, but correct me if I m wrong, that with the two available authentication mechanisms we will face the following issues:
- Client Credentials: With this approach we won't be able to get the files property from API's responses as we are not logged in.
- Code Grant: It is not possible to launch Safari on tvOS ([no WebKIt and UIWebViews either](http://www.wired.com/2015/09/apple-tv-web/)) plus we want the app to be logged in as our PRO account (to retrieve the list of our videos). Users won't know our credentials nor we want to share them.

While going through the VIMNetworking repo I understood that what we are trying to achieve is referred as [Lightweight use](https://github.com/vimeo/VIMNetworking#lightweight-use)

Now going back to the first part of your comment, I also feel that shielding the ```currentAccount``` from other modules is the best approach but we still need some ways to use the Bearer generated for API apps. 

One possible solution would be to add a public setter function for the Bearer in the ```AuthenticationController``` that will ultimately set the ```currentAccount``` property of the VimeoClient.

What do you think?

huebnerob:
Sorry for the delayed response @nicolasabric, I was away last week. 

Yeah, I see the issue here, we need a way for VimeoNetworking to accept a constant token.  I agree with your last statement, which is that `AuthenticationController` should have a function to do this.  What do you think about adding something like an `accessToken(token: String)` method to the controller?  Note that you can make an api request to **/oauth/verify** to get a `VIMAccount` object when you already provide an access token in the response

nicolasabric:
@huebnerob don't worry. I will update the PR and integrate your comments.

nicolasabric:
@huebnerob as requested I have created a new function in the ```AuthenticationController```. Signature is a bit different than the one you suggested as I have added a completion block. The approach I took is to consider constant access token as yet another authentication method. Therefore the ```accessToken(token: String, completion: AuthenticationCompletion)``` behave as follow for a given token:
- Create an adhoc ```VimeoClient``` with a custom ```accessTokenProvider``` 
- Create an ```AuthenticationRequest``` to **oauth/verify**.
- Call ```authenticate``` function on the adhoc ```VimeoClient```. 

From there normal authentication flow resumes and external VimeoClient's ```currentAccount``` is finally set with the response of the oauth/verify call.

Now you may wonder why I did not used the ```authenticatorClient``` and instead created an adhoc one. From what I understood ```authenticatorClient``` should remain unauthenticated and setting the constant token on it would have changed. 

Do not hesitate to comment/correct me if I m wrong.

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -227,6 +227,20 @@ final public class AuthenticationController
         self.authenticate(request: request, completion: completion)
     }
     
+    /**
+     Execute a constant token grant request. This type of authentication allows access to public and personnal content on Vimeo. Constant token are usually generated for API apps see https://developer.vimeo.com/apps
+     
+     - parameter completion: handles authentication success or failure
```
huebnerob: Can we document the `token` parameter here as well?

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -227,6 +227,20 @@ final public class AuthenticationController
...
+     */
+    public func accessToken(token: String, completion: AuthenticationCompletion)
+    {
+        let customSessionManager =  VimeoSessionManager.defaultSessionManager(accessTokenProvider: {() -> String? in return "\(token)"})
```
huebnerob: Small thing, we can actually simplify

`{() -> String? in return "\(token)"}`

to just

`{ token }`

huebnerob:
This looks great @nicolasabric, thanks!  Left two small comments, but I'm good to merge after those

nicolasabric:
@huebnerob I have done the modifications and also added a Lightweight Use section in the Readme.md to document the "new" authentication method.

huebnerob:
Awesome, thanks @nicolasabric! Merging...

PR #39, opened 2016-07-05
CAM-XXX: Fixed a bug in the NSError extension parsing logic

mikew-personal:
#### Ticket
N/A

#### Ticket Summary
Fixed a bug in the NSError extension parsing logic for the invalid_parameters array returned in the error JSON object.

#### Implementation Summary
In the vimeoInvalidParametersErrorCodes property we were incorrectly casting the value of the "error_code" field to a String and then attempting to cast it to an Int before adding it to the returned array.  The value was already an Int in the JSON data, so the cast to string was failing which caused us to always return an empty array.  I fixed the logic to remove the incorrect String cast.  I also added a set of unit tests to make sure all of the properties in the extension are returning the correct values.  This commit also includes a new convenience property to get the VimeoUserMessage from the invalid parameters error JSON object.

#### How to Test
1. Run the unit tests.

### DIFF
--- a/VimeoNetworking/VimeoNetworking/NSError+Extensions.swift
+++ b/VimeoNetworking/VimeoNetworking/NSError+Extensions.swift
@@ -247,11 +247,8 @@ public extension NSError
         }
         
         for errorJSON in invalidParameters {
-            if let errorCode = errorJSON[self.dynamicType.VimeoErrorCodeKey] as? String {
-                
-                if let errorCodeInt = Int(errorCode) {
-                    errorCodes.append(errorCodeInt)
-                }
+            if let errorCode = errorJSON[self.dynamicType.VimeoErrorCodeKey] as? Int {
+                errorCodes.append(errorCode)
             }
         }
         
@@ -262,6 +259,17 @@ public extension NSError
     public var vimeoInvalidParametersFirstErrorCode: Int
     {
         return self.vimeoInvalidParametersErrorCodes.first ?? NSNotFound
+    }
+    
+    public var vimeoInvalidParametersFirstVimeoUserMessage: String? {
+        
+        guard let json = self.errorResponseBodyJSON,
+            invalidParameters = json[self.dynamicType.VimeoInvalidParametersKey] as? NSArray else {
+            return nil
+        }
+        
+        let errorJSON = invalidParameters.firstObject
+        return errorJSON?[self.dynamicType.VimeoUserMessageKey] as? String
     }
     
         /// Returns an underscore separated string of all the error codes in the the api error JSON dictionary if any exist, otherwise returns nil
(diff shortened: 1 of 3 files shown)


VimeoNetworking/VimeoNetworking/NSError+Extensions.swift
```diff
@@ -247,11 +247,8 @@ public extension NSError
         }
         
         for errorJSON in invalidParameters {
-            if let errorCode = errorJSON[self.dynamicType.VimeoErrorCodeKey] as? String {
```
huebnerob: Are we sure this doesn't break for any error codes including upload and authentication? Dunno why it would, but I know this code has existed as part of VimeoUpload for a while. cc @alfiehanssen

VimeoNetworking/VimeoNetworkingTests/NSErrorExtensionTests.swift
```diff
@@ -0,0 +1,91 @@
...
+//  NSErrorExtensionTests.swift
+//  VimeoNetworking
+//
+//  Created by Westendorf, Michael on 7/5/16.
+//  Copyright © 2016 Vimeo. All rights reserved.
+//
+
+import XCTest
+
+class NSErrorExtensionTests: XCTestCase {
```
huebnerob: This is great!

huebnerob:
Looks good @mikew-personal.  As long as we're sure this doesn't mess with upload errors.

VimeoNetworking/VimeoNetworkingTests/NSErrorExtensionTests.swift
```diff
@@ -60,11 +60,28 @@ class NSErrorExtensionTests: XCTestCase {
...
     
-    func testVimeoServerErrorCode() {
+    func testVimeoServerErrorCodeInJSONErrorBody() {
         XCTAssertNotNil(self.testApiError.vimeoServerErrorCode)
         XCTAssertEqual(self.testApiError.vimeoServerErrorCode!, 2204)
     }
     
+    func testLegacyVimeoServerErrorCode() {
+        var userInfo = self.testApiError.userInfo
+        userInfo["VimeoErrorCode"] = 4101
```
huebnerob: Isn't there a `VimeoErrorCode` constant somewhere?

huebnerob:
@mikew-personal 👍

VimeoNetworking/VimeoNetworkingTests/NSErrorExtensionTests.swift
mikew-personal: it's a private constant in the NSError+Extensions.swift file.  I don't know if the @testable keyword works with class extensions.

PR #42, opened 2016-07-18
Adding cache removal method

huebnerob:
#### Ticket
[TVOS-14](https://vimean.atlassian.net/browse/TVOS-14)

#### Ticket Summary
Port stream data source logic from iOS to tvOS. This will eliminate VIMNetworking from our stream logic and replace it with VimeoNetworking.

#### Implementation Summary
- Added remove cache for `Request` method

#### How to Test
See app-side PRs

### DIFF
--- a/VimeoNetworking/VimeoNetworking/VimeoClient.swift
+++ b/VimeoNetworking/VimeoNetworking/VimeoClient.swift
@@ -279,7 +279,12 @@ final public class VimeoClient
         return RequestToken(task: requestTask)
     }
     
-    public func removeCachedResponse<ModelType: MappableResponse>(request request: Request<ModelType>)
+    /**
+     Removes any cached responses for a given `Request`
+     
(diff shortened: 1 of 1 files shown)


VimeoNetworking/VimeoNetworking/VimeoClient.swift
```diff
@@ -279,6 +279,11 @@ final public class VimeoClient
         return RequestToken(task: requestTask)
     }
     
+    public func removeCachedResponse<ModelType: MappableResponse>(request request: Request<ModelType>)
```
jasonhawkins: This took me a minute to unpack, but I'm reading this as, "remove the cached response of a generic model type, which is any `Request` object that conforms to `MappableResponse`." Does that sound right?

Sometimes I think Swift is not the easiest to read.

jasonhawkins:
👍🏼

PR #43, opened 2016-07-25
Adding associatedtype Element to MappableResponse 

huebnerob:
This fixes an issue where we couldn't cast an Array response to an expected array of model objects correctly.  By adding the associated type, it allows us to restrict the generic parameters in a function or class as such: `<ModelType: MappableResponse where ModelType.Element: VIMModelObject>`.  This merely makes explicit what we already know, that we're only requesting model objects from the api, and it also fixes our ability to make that cast: `let newItems = response.model as? [ModelType.Element]`


mikew-personal:
👍 Looks good.

PR #44, opened 2016-07-28
TVOS-228: Row Streams View Controller on tvOS

huebnerob:
#### Ticket
[https://vimean.atlassian.net/browse/TVOS-228](https://vimean.atlassian.net/browse/TVOS-228)

#### Ticket Summary
Implement a Streams view controller that can present and scroll sections as rows individually

#### Implementation Summary

**see Vimeo-tvOS PR for context on the user-facing feature** 

Modified two things relating to caching:
* cache keys now all start with "cached" (for aesthetics really) and they omit the "fields" parameter because it was making our filenames too long
* `ResponseCache` no longer generates an actual `Response` object from the response dictionary.  It instead passes control of this dictionary back to `VimeoClient` for parsing into the final `Response` object.  This is to allow the parsing of pagination parameters to stay contained in `VimeoClient`, but it's also nice to consolidate this architecturally.

### DIFF
--- a/VimeoNetworking/VimeoNetworking/VimeoClient.swift
+++ b/VimeoNetworking/VimeoNetworking/VimeoClient.swift
@@ -176,14 +176,11 @@ final public class VimeoClient
                 
                 switch result
                 {
-                case .Success(let response):
+                case .Success(let responseDictionary):
                     
-                    if let response = response
+                    if let responseDictionary = responseDictionary
                     {
-                        dispatch_async(completionQueue)
-                        {
-                            completion(result: .Success(result: response))
-                        }
+                        self.handleTaskSuccess(request: request, task: nil, responseObject: responseDictionary, isCachedResponse: true, isFinalResponse: request.cacheFetchPolicy == .CacheOnly, completionQueue: completionQueue, completion: completion)
                     }
                     else if request.cacheFetchPolicy == .CacheOnly
                     {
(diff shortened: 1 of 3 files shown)


jasonhawkins:
> This is to allow the parsing of pagination parameters to stay contained in VimeoClient, but it's also nice to consolidate this architecturally.

This sounds like a great idea ✨

VimeoNetworking/VimeoNetworking/VimeoClient.swift
```diff
@@ -388,6 +387,8 @@ final public class VimeoClient
         }
         catch let error
         {
+            self.responseCache.removeResponseForRequest(request)
```
jasonhawkins: Do you think we'd ever _not_ need to remove the cached response when we have an error? From here it looks like regardless of what the error actually is we'll just dump the cache for this request. Or is that fine?

jasonhawkins:
Just had the one question buuuut I think you're pretty solid here🏌

VimeoNetworking/VimeoNetworking/VimeoClient.swift
huebnerob: One possible error that may want to hold on to the cache is something like a 'No Internet' error, but I believe we're already handling that upstream of this logic in that we don't even attempt to make a request when we have no connection.  I can't really think of any other error at the moment that would fall into the same pattern.

PR #45, opened 2016-08-08
TVOS-85: Feed Stream

huebnerob:
#### Ticket
[TVOS-85](https://vimean.atlassian.net/browse/TVOS-85)
[TVOS-81](https://vimean.atlassian.net/browse/TVOS-81)

#### Implementation Summary
* Tweaked the ways next page `Request`s are generated and accessed for more consistency.
* When you init a `Request` with a `path`, it will now strip any query string parameters correctly and put them in the parameters and pagination variables as appropriate.
* You can now get a `Request.URI`, which is the path plus a query string of any parameters. This simplifies stream next page requests.

### DIFF
--- a/VimeoNetworking/VimeoNetworking/Request.swift
+++ b/VimeoNetworking/VimeoNetworking/Request.swift
@@ -186,8 +186,8 @@ public struct Request<ModelType: MappableResponse>
             }
             self.itemsPerPage = itemsPerPage
             
-            var additionalParameters: VimeoClient.RequestParameters = queryParameters
-            parameters?.forEach { (key, object) in
+            var additionalParameters = parameters ?? [:]
+            queryParameters.forEach { (key, object) in
                 additionalParameters[key] = object
             }
             self.additionalParameters = additionalParameters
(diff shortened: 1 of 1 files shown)


VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -164,31 +165,69 @@ public struct Request<ModelType: MappableResponse>
...
+        let (path, query) = path.splitLinkString()
         self.path = path
-        self.additionalParameters = parameters ?? [:]
-        self.page = page
-        self.itemsPerPage = itemsPerPage
+        
+        if let query = query,
+            let queryParameters = query.parametersFromQueryString()
+        {
+            var page = page
```
jasonhawkins: These `if let`s look similar enough that I wonder if we could come up with some kind of function that uses generics to do this. Does that seem like it could be a thing? They're all trying to conditionally set a variable and assigning it to a property of the same type.

huebnerob: Maybe possibly, but it would be tough because swift doesn't have runtime introspection so you can't go from `"page"` to `self.page` in code.  

These are good thoughts to funnel into a picture of what a swift JSON model object parser might look like though.  Swift 4 is already talking about better introspective features, and I for one can't wait 😁

PR #46, opened 2016-08-12
Dispatch VimeoClient task success/failure to a background queue for parsing

huebnerob:
Incorrectly assumed that AFNetworking returned us to a background queue in their completion blocks, but they actually default to the main queue, so all our model object parsing and whatnot was being done there 😱


huebnerob:
@mikew-personal bump 🐛

alfiehanssen:
Looks great @huebnerob. One question, does AF allow us to specify a callback queue and would this be simpler / faster to use?

huebnerob:
@alfiehanssen alas, it does not, unfortunately

PR #53, opened 2016-09-09
Auto xcode changes.

alfiehanssen:


alfiehanssen:
@huebnerob @mikew-personal Re-ping...

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
```diff
@@ -166,7 +166,7 @@ final public class AuthenticationController
                           self.dynamicType.ScopeKey: Scope.combine(self.configuration.scopes),
                           self.dynamicType.StateKey: self.dynamicType.state]
         
-        guard let urlString = VimeoBaseURLString?.URLByAppendingPathComponent(self.dynamicType.CodeGrantAuthorizationPath).absoluteString
+        guard let urlString = VimeoBaseURLString?.URLByAppendingPathComponent(self.dynamicType.CodeGrantAuthorizationPath)!.absoluteString
```
huebnerob: why can't we just make this optional if it's in a guard let?

huebnerob: I mean, keep it optionally unwrapped

huebnerob:
:

VimeoNetworking/VimeoNetworking/ResponseCache.swift
```diff
@@ -293,7 +293,7 @@ final internal class ResponseCache
             
             let fileURL = directoryURL.URLByAppendingPathComponent(key)
```
mikew-personal: Can we use a guard let here for fileURL and use fatalError() if fileURL is nil (similar to what cachesDirectoryURL() does above).  That way we can get a more descriptive error instead of a forced unwrap crash if fileURL does end up being nil.

huebnerob:
@alfiehanssen Do we still need to merge this?

VimeoNetworking/VimeoNetworking/AuthenticationController.swift
alfiehanssen: Yes, this was an auto-Xcode change, I'm revising now

PR #56, opened 2016-09-20
VIM-4562 Onboarding Recommendations Model

nicolelehrer:
#### Ticket
[VIM-4562](https://vimean.atlassian.net/browse/VIM-4562)

#### Ticket Summary
Added model support for Recommendations. See what a recommendation response looks like here: https://vimean.atlassian.net/wiki/display/DEV/Onboarding+Spec

#### Implementation Summary
- added recommendations connection keys 
- added model class that will either populate its category or user object based on the recommendation type 
 
#### How to Test
N/A See parent ticket

### DIFF
--- a/VimeoNetworking.podspec
+++ b/VimeoNetworking.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name         = "VimeoNetworking"
-  s.version      = "0.0.1"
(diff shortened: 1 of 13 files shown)


nicolelehrer:
This was branched off the wrong branch - 0.0.2 instead of master. I reverted the commits that were different from master so only my new changes will be merged into master.

PR #60, opened 2016-09-24
Resolve compatibility issues with iOS project

nicolelehrer:
#### Ticket
NA

#### Ticket Summary
The iOS project does not support `VimeoNetworking` as a Cocoapod yet, but this is forthcoming. The TVOS project does, and needs to import this header file to make a new model written in swift visible to `VIMVideo` written in Objective-C. 

#### Implementation Summary
Wrap import in macro to check OS.

#### How to Test
See parent PR.


nicolelehrer:
thanks @huebnerob !

PR #71, opened 2016-10-29
Added MIT License Header

alfiehanssen:
Added license header to all files. Updated module name that appears in comments above license header.

Many files touched, worth a glance but the changes are strictly copying and pasting the license header into place, and modifying the module name in comments just above the license header.


alfiehanssen:
@ghking Hit me with a retroactive review please

PR #72, opened 2016-11-01
VIM-4736: AnyObject Parameters

ghking:
#### Ticket

https://vimean.atlassian.net/browse/VIM-4736

#### Ticket Summary

Previously, VimeoNetworking only supported dictionary types for request parameters. We need to support array types as well as required by certain requests to out API.

#### Implementation Summary

- Modified the `parameters` of Request to be of type AnyObject (corresponding to the AFNetworking and Apple APIs)
- Cleaned up some cache code
- Moved some request code from the app to VimeoNetworking

#### How to Test

Make sure networking works as expected.

### DIFF
--- a/VimeoNetworking/VimeoNetworking/Request+Authentication.swift
+++ b/VimeoNetworking/VimeoNetworking/Request+Authentication.swift
@@ -55,8 +55,8 @@ extension Request where ModelType: VIMAccount
      */
     static func clientCredentialsGrantRequest(scopes scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypeClientCredentials,
-                                                         ScopeKey: Scope.combine(scopes)]
+        let parameters = [GrantTypeKey: GrantTypeClientCredentials,
+                          ScopeKey: Scope.combine(scopes)]
         
         return Request(method: .POST, path: AuthenticationPathClientCredentials, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -71,9 +71,9 @@ extension Request where ModelType: VIMAccount
      */
     static func codeGrantRequest(code code: String, redirectURI: String) -> Request
     {
-        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypeAuthorizationCode,
-                                                         CodeKey: code,
-                                                         RedirectURIKey: redirectURI]
+        let parameters = [GrantTypeKey: GrantTypeAuthorizationCode,
+                          CodeKey: code,
+                          RedirectURIKey: redirectURI]
         
         return Request(method: .POST, path: AuthenticationPathCodeGrant, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -89,10 +89,10 @@ extension Request where ModelType: VIMAccount
      */
     static func logInRequest(email email: String, password: String, scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypePassword,
-                                                         ScopeKey: Scope.combine(scopes),
-                                                         UsernameKey: email,
-                                                         PasswordKey: password]
+        let parameters = [GrantTypeKey: GrantTypePassword,
+                          ScopeKey: Scope.combine(scopes),
+                          UsernameKey: email,
+                          PasswordKey: password]
         
         return Request(method: .POST, path: AuthenticationPathAccessToken, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
--- a/VimeoNetworking/VimeoNetworking/Request+Authentication.swift
+++ b/VimeoNetworking/VimeoNetworking/Request+Authentication.swift
@@ -55,7 +55,7 @@ extension Request where ModelType: VIMAccount
      */
     static func clientCredentialsGrantRequest(scopes scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypeClientCredentials,
+        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypeClientCredentials,
                                                          ScopeKey: Scope.combine(scopes)]
         
         return Request(method: .POST, path: AuthenticationPathClientCredentials, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
@@ -71,7 +71,7 @@ extension Request where ModelType: VIMAccount
      */
     static func codeGrantRequest(code code: String, redirectURI: String) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypeAuthorizationCode,
+        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypeAuthorizationCode,
                                                          CodeKey: code,
                                                          RedirectURIKey: redirectURI]
         
@@ -89,7 +89,7 @@ extension Request where ModelType: VIMAccount
      */
     static func logInRequest(email email: String, password: String, scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypePassword,
+        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypePassword,
                                                          ScopeKey: Scope.combine(scopes),
                                                          UsernameKey: email,
                                                          PasswordKey: password]
@@ -104,7 +104,7 @@ extension Request where ModelType: VIMAccount
      */
     static func verifyAccessTokenRequest() -> Request
     {
-        return Request(method: .GET, path: AuthenticationPathAccessTokenVerify, parameters: nil, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
+        return Request(method: .GET, path: AuthenticationPathAccessTokenVerify, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
     
     /**
@@ -119,10 +119,10 @@ extension Request where ModelType: VIMAccount
      */
     static func joinRequest(name name: String, email: String, password: String, scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [ScopeKey: Scope.combine(scopes),
-                                                         DisplayNameKey: name,
-                                                         EmailKey: email,
-                                                         PasswordKey: password]
+        let parameters = [ScopeKey: Scope.combine(scopes),
+                          DisplayNameKey: name,
+                          EmailKey: email,
+                          PasswordKey: password]
         
         return Request(method: .POST, path: AuthenticationPathUsers, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -137,9 +137,9 @@ extension Request where ModelType: VIMAccount
      */
     static func logInFacebookRequest(facebookToken facebookToken: String, scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypeFacebook,
-                                                         ScopeKey: Scope.combine(scopes),
-                                                         TokenKey: facebookToken]
+        let parameters = [GrantTypeKey: GrantTypeFacebook,
+                          ScopeKey: Scope.combine(scopes),
+                          TokenKey: facebookToken]
         
         return Request(method: .POST, path: AuthenticationPathFacebookToken, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -154,8 +154,8 @@ extension Request where ModelType: VIMAccount
      */
     static func joinFacebookRequest(facebookToken facebookToken: String, scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [ScopeKey: Scope.combine(scopes),
-                                                         TokenKey: facebookToken]
+        let parameters = [ScopeKey: Scope.combine(scopes),
+                          TokenKey: facebookToken]
         
         return Request(method: .POST, path: AuthenticationPathUsers, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -170,8 +170,8 @@ extension Request where ModelType: VIMAccount
      */
     static func authorizePinCodeRequest(userCode userCode: String, deviceCode: String) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [PinCodeKey: userCode,
-                                                         DeviceCodeKey: deviceCode]
+        let parameters = [PinCodeKey: userCode,
+                          DeviceCodeKey: deviceCode]
         
         return Request(method: .POST, path: AuthenticationPathPinCodeAuthorize, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -185,7 +185,7 @@ extension Request where ModelType: VIMAccount
      */
     static func appTokenExchangeRequest(accessToken accessToken: String) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [AccessTokenKey: accessToken]
+        let parameters = [AccessTokenKey: accessToken]
         
         return Request(method: .POST, path: AuthenticationPathAppTokenExchange, parameters: parameters, cacheFetchPolicy: .NetworkOnly, shouldCacheResponse: false)
     }
@@ -201,6 +201,20 @@ extension Request where ModelType: VIMNullResponse
     public static func deleteTokensRequest() -> Request
     {
         return Request(method: .DELETE, path: AuthenticationPathTokens, retryPolicy: .TryThreeTimes)
+    }
+    
+    /**
+     Construct a `Request` for resetting the user's password
+     
+     - parameter email: the user's email
+     
+     - returns: a new `Request`
+     */
+    public static func resetPasswordRequest(withEmail email: String) -> Request
+    {
+        let path = "/users/" + email + "/password/reset"
+        
+        return Request(method: .POST, path: path)
     }
 }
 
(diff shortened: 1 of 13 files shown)


VimeoNetworking/VimeoNetworking/Request+Authentication.swift
```diff
@@ -55,7 +55,7 @@ extension Request where ModelType: VIMAccount
      */
     static func clientCredentialsGrantRequest(scopes scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypeClientCredentials,
+        let parameters: VimeoClient.RequestDictionary = [GrantTypeKey: GrantTypeClientCredentials,
```
alfiehanssen: Per my comment in the other PR, maybe `RequestParametersDictionary?

VimeoNetworking/VimeoNetworking/Request+Authentication.swift
```diff
@@ -219,9 +233,9 @@ extension Request where ModelType: PinCodeInfo
      */
     static func getPinCodeRequest(scopes scopes: [Scope]) -> Request
     {
-        let parameters: VimeoClient.RequestParameters = [GrantTypeKey: GrantTypePinCode,
```
alfiehanssen: What is the advantage / disadvantage of having this `VimeoClient.RequestParameters` type specified here?

VimeoNetworking/VimeoNetworking/Request+Cache.swift
```diff
@@ -8,27 +8,21 @@
...
         {
-            if key == "fields"
-            {
-                // avoiding the field filtering key due to excessive length
-                continue
-            }
-            
-            cacheKey += key
-            cacheKey += value.description
+            cacheKey = cacheKey + "." + String(parameters.description.hashValue)
```
alfiehanssen: Nice

VimeoNetworking/VimeoNetworking/Request+Configs.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  Request+Configs.swift
+//  Pods
+//
+//  Created by King, Gavin on 10/31/16.
+//
```
alfiehanssen: Needs license header

VimeoNetworking/VimeoNetworking/Request+Video.swift
```diff
@@ -34,6 +34,49 @@ public extension Request
...
+    /**
+     Create a `Request` to get a specific VOD video
+     
+     - parameter vodVideoURI: the VOD video's URI
+     
+     - returns: a new `Request`
+     */
+    static func vodVideoRequest(vodVideoURI vodVideoURI: String) -> Request
+    {
+        let parameters = ["_video_override": "true"]
```
alfiehanssen: Ugh, what is up with this: `_video_override`, blargh

VimeoNetworking/VimeoNetworking/Request+Video.swift
```diff
@@ -34,6 +34,49 @@ public extension Request
...
+    /**
+     Create a `Request` to get the selected users for the vieo
+     
+     - parameter videoURI: the URI of the video
+     
+     - returns: a new `Request`
+     */
+    static func selectedUsersRequest(videoURI videoURI: String) -> Request
+    {
+        let parameters = [VimeoClient.PerPageKey: 100]
```
alfiehanssen: Break out constant(s)?

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -99,43 +99,20 @@ public struct Request<ModelType: MappableResponse>
         /// request url path (e.g. `/me`, `/videos/123456`)
     public let path: String
     
-        /// any parameters to include with the request, this is calculated from the raw parameters provided at request initialization plus any specified `page` or `itemsPerPage` values on the request itself
-    public var parameters: VimeoClient.RequestParameters
```
alfiehanssen: Why are we able to remove this logic? Is there benefit to keeping it?

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -155,50 +130,16 @@ public struct Request<ModelType: MappableResponse>
...
-        
-        if let query = query,
-            let queryParameters = query.parametersFromQueryString()
-        {
-            var page = page
-            if let pageValue = queryParameters[self.PageKey]
-            {
-                page = page ?? Int(pageValue)
-            }
-            self.page = page
```
alfiehanssen: Does removing this limit our ability to inspect the page and items per page values on the request?

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -210,10 +151,13 @@ public struct Request<ModelType: MappableResponse>
     {
         var URI = self.path
         
-        let queryString = AFQueryStringFromParameters(self.parameters)
-        if queryString.characters.count > 0
+        if let parameters = self.parameters as? VimeoClient.RequestDictionary
```
alfiehanssen: Where is the array factored into things?

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -210,10 +151,13 @@ public struct Request<ModelType: MappableResponse>
...
         {
-            URI += "?" + queryString
+            let queryString = AFQueryStringFromParameters(parameters)
+            if queryString.characters.count > 0
+            {
+                URI += "?" + queryString
```
alfiehanssen: I think we can/should be using NSURLComponents to construct the URL, avoids manually assembling a string...

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -223,14 +167,37 @@ public struct Request<ModelType: MappableResponse>
...
                        cacheFetchPolicy: self.cacheFetchPolicy,
                        shouldCacheResponse: self.shouldCacheResponse,
                        retryPolicy: self.retryPolicy)
     }
-}
\ No newline at end of file
+    
+    private static func parametersByRemovingQueryParameters(inPath path: String, parameters: AnyObject?) -> AnyObject?
+    {
+        guard var parametersDictionary = parameters as? VimeoClient.RequestDictionary else
```
alfiehanssen: I'm a little confused, it seems we're doing a nil check as well as a type check here. But we don't do anything with the `parameters` argument below. So can we just have the argument type match the type we're checking for here?

alfiehanssen: Since we're returning a dictionary we should be able to specify a return type of dictionary, right?

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -223,14 +167,37 @@ public struct Request<ModelType: MappableResponse>
...
+                       parameters: parameters,
                        modelKeyPath: self.modelKeyPath,
                        cacheFetchPolicy: self.cacheFetchPolicy,
                        shouldCacheResponse: self.shouldCacheResponse,
                        retryPolicy: self.retryPolicy)
     }
-}
\ No newline at end of file
+    
+    private static func parametersByRemovingQueryParameters(inPath path: String, parameters: AnyObject?) -> AnyObject?
```
alfiehanssen: Do we want to remove query params and place them in the request body? It seems like this might be confusing to consumers / users of the library.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
```diff
@@ -64,24 +73,24 @@ final internal class ResponseCache
     {
         let key = request.cacheKey
         
-        self.memoryCache.removeResponseDictionaryForKey(key)
```
alfiehanssen: Consider keeping the method names more simple / generic. I.e. dropping the term "response dictionary" and just using "response"

VimeoNetworking/VimeoNetworking/ResponseCache.swift
```diff
@@ -90,19 +99,19 @@ final internal class ResponseCache
             self.cache.setObject(responseDictionary, forKey: key)
         }
         
-        private func responseDictionaryForKey(key: String) -> VimeoClient.ResponseDictionary?
+        private func responseDictionary(forKey key: String, completion: ResponseDictionaryClosure)
```
alfiehanssen: Why not use a return value instead of a closure? Return value is logically simpler and there doesn't seem to be an async aspect to this.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
```diff
@@ -8,6 +8,19 @@
 
 import Foundation
 
+private typealias ResponseDictionaryClosure = (responseDictionary: VimeoClient.ResponseDictionary?) -> Void
+
+private protocol Cache
```
alfiehanssen: Could this just be generic? This would mean that we don't have to specify the return type here, and mean that the method names wouldn't have to state the type.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
```diff
@@ -35,22 +48,18 @@ final internal class ResponseCache
...
-        {
-            completion(result: .Success(result: responseDictionary))
-        }
-        else
-        {
-            self.diskCache.responseDictionaryForKey(key) { responseDictionary in
-                
-                if let responseDictionary = responseDictionary
-                {
+        self.memoryCache.responseDictionary(forKey: key) { (responseDictionary) in
```
alfiehanssen: I don't think that memory cache should have a callback, should be return value imo

alfiehanssen: I don't think standardizing the protocol is worth the cost of making this a closure instead of a return val

VimeoNetworking/VimeoNetworking/String+Parameters.swift
```diff
@@ -15,9 +15,9 @@ public extension String
      
      - returns: a dictionary of parameters if parsing succeeded
      */
-    func parametersFromQueryString() -> [String: String]?
+    func parametersDictionaryFromQueryString() -> [String: String]?
```
alfiehanssen: Where does the array body come into play?

VimeoNetworking/VimeoNetworking/VimeoClient.swift
```diff
@@ -55,7 +55,10 @@ final public class VimeoClient
     }
     
         /// Dictionary containing URL parameters for a request
-    public typealias RequestParameters = [String: AnyObject]
+    public typealias RequestDictionary = [String: AnyObject]
+    
+        /// Array containing URL parameters for a request
+    public typealias RequestArray = [AnyObject]
```
alfiehanssen: Can this type be more communicative in its name. Sounds like an array of requests. Maybe something like RequestBodyArray

VimeoNetworking/VimeoNetworking/VimeoClient.swift
```diff
@@ -55,7 +55,10 @@ final public class VimeoClient
     }
     
         /// Dictionary containing URL parameters for a request
-    public typealias RequestParameters = [String: AnyObject]
+    public typealias RequestDictionary = [String: AnyObject]
```
alfiehanssen: On that note (below) this name sounds kinda like a dictionary of requests. I think a name like RequestParameterDictionary would satisfy both concerns. A little longer but not bad.

VimeoNetworking/VimeoNetworking/VimeoClient.swift
```diff
@@ -65,14 +68,14 @@ final public class VimeoClient
...
     
-    private static let PagingKey = "paging"
-    private static let TotalKey = "total"
-    private static let PageKey = "page"
-    private static let PerPageKey = "per_page"
-    private static let NextKey = "next"
-    private static let PreviousKey = "previous"
-    private static let FirstKey = "first"
-    private static let LastKey = "last"
+    internal static let PagingKey = "paging"
```
alfiehanssen: Is it possible to keep these private?

alfiehanssen:
👍  Done with review, interested to hear your perspective on some of the questions etc.

VimeoNetworking/VimeoNetworking/Request+Authentication.swift
ghking: Yeah I'm down with that

VimeoNetworking/VimeoNetworking/Request+Authentication.swift
ghking: It's unnecessary with type inference

VimeoNetworking/VimeoNetworking/Request.swift
ghking: Nope, no longer necessary. I'm removing pagination info from the Request class completely since not every (almost all) requests don't have anything to do with pagination (see my next comment).

VimeoNetworking/VimeoNetworking/Request.swift
ghking: Probably. This was exiting copy and pasted logic though so I'm going to keep it as is for now

VimeoNetworking/VimeoNetworking/Request.swift
ghking: Only if we change the argument specifically be a dictionary as well. See the above comment ^

VimeoNetworking/VimeoNetworking/Request.swift
ghking: Yeah good call. I picked this way since its easier to mutate a dictionary that a URL, but that makes more sense. but I'll modify this.

VimeoNetworking/VimeoNetworking/Request.swift
```diff
@@ -223,14 +167,37 @@ public struct Request<ModelType: MappableResponse>
     
     internal func associatedPageRequest(newPath newPath: String) -> Request<ModelType>
     {
+        // Since page response paging paths bake the paging parameters into the path (even if we originally provide them in the body),
```
ghking: See the below comment. I'm going to modify this to keep the body intact and update the paging values with the updated values provided in the query params rather than keeping the url intact and deleting duplicate params in the body.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
ghking: I'm using the "responseDictionary" name since we're dealing with dictionaries rather than Responses. It might be weird if we're telling it to remove a "response" which is actually a dictionary and not a Response. Thoughts?

VimeoNetworking/VimeoNetworking/ResponseCache.swift
ghking: Agreed, but this function is an implementation of a function in the `Cache` protocol. While this memory cache isn't async, the disk cache is. I'd rather return the response in a block in the memory cache and have the two caches follow a shared contract than have different getter functions for each.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
ghking: Possibly, that's probably outside the scope of this ticket though. I just quickly added this protocol (already outside of the scope) since I noticed the caches we're sharing the same function names.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
ghking: I really like the idea of having `Cache`s which have a shared, structured contract. This is just syntax which has no functional impact. I can remove the protocol if you feel strongly about it though.

VimeoNetworking/VimeoNetworking/VimeoClient.swift
ghking: Agreed. Gonna change these guys to "RequestParametersDictionary" and "RequestParametersArray"

VimeoNetworking/VimeoNetworking/VimeoClient.swift
ghking: see above

VimeoNetworking/VimeoNetworking/VimeoClient.swift
ghking: internal is the best we can get since we use these guys outside of VimeoClient

VimeoNetworking/VimeoNetworking/Request.swift
alfiehanssen: Pretty please it will take like 3 minutes 😁

VimeoNetworking/VimeoNetworking/Request.swift
alfiehanssen: Okay, I suggest adding this additional info in a comment in this file. Most people will want this same background info. ❤️

VimeoNetworking/VimeoNetworking/Request.swift
alfiehanssen: Any opportunity we have to make things more clear, explicit etc I think we should take. Seems like there's some room for improvement in that area here.

alfiehanssen: Yeah, what I'm saying is that it seems like we can safely just call it a dictionary everywhere (and eliminate AnyObject), unless I'm missing something.

VimeoNetworking/VimeoNetworking/ResponseCache.swift
alfiehanssen: Gotcha, nit: wondering if this should not just be setValueForKey, valueForKey, removeValueForKey etc?

VimeoNetworking/VimeoNetworking/ResponseCache.swift
alfiehanssen: I think the cost of using a closure instead of a return value for the memory cache is greater than the cost of either not using or of modifying the cache protocol.

alfiehanssen: Can you make the closure function optional and introduce a return val function that's also optional. Or do something to enforce one or the other. Or just implement both?

VimeoNetworking/VimeoNetworking/ResponseCache.swift
alfiehanssen: I kinda think that if the cache work was outside then by doing it you kinda are obligated to make it maximally nice 🤓

VimeoNetworking/VimeoNetworking/ResponseCache.swift
alfiehanssen: I just don't like the idea of a closure when a return value is more appropriate.

VimeoNetworking/VimeoNetworking/Request.swift
ghking: mk fine

VimeoNetworking/VimeoNetworking/Request.swift
ghking: Done. Also there's already a comment but I'll update it with the new strategy

VimeoNetworking/VimeoNetworking/Request+Configs.swift
ghking: Gonna add these after the merge

PR #80, opened 2016-11-28
tvOS-475: Improve DRM flagging

jasonhawkins:
#### Ticket
[tvOS-475](https://vimean.atlassian.net/browse/tvOS-475)

#### Ticket Summary
tvOS-475 requires that we handle available actions for DRM-protected content different from other VOD items. 

#### Implementation Summary
- **VIMInteraction:** – Added a `drm` property to capture the DRM status from the API. 
- **VIMVideo:** – Changed the implementation of `-isDRMProtected` to check the status of the `drm` BOOL instead of looking for FairPlay files in the `drmFiles` array. Checking the `drm` property instead of the array gives us direct information about whether the content is DRM-protected, instead of indirectly relying on the array.
- **VIMSeason:** – Added to VimeoNetworking project and implemented `-isDRMProtected`.
- Sorted _Models_ folder by name.

#### How to Test
n/a


VimeoNetworking/VimeoNetworking/Models/VIMInteraction.h
```diff
@@ -55,4 +55,11 @@ typedef NS_ENUM(NSInteger, VIMInteractionStreamStatus) {
 @property (nonatomic, strong, nullable) NSDate *purchaseDate;
 @property (nonatomic, assign) VIMInteractionStreamStatus streamStatus;
 
+# pragma mark - DRM
+
+/**
+ Indicates whether this VIMInteraction is protected by DRM, which applies to items available for rent or purchase.
+ */
+@property (nonatomic, assign) BOOL drm;
```
ghking: How can an interaction be protected by DRM? An interaction is not content.

ghking: Also maybe isDRM since its a predicate

ghking: Or to address both issues maybe something more descriptive that speaks more to the fact that the result of the interaction will lead to DRM protected content

VimeoNetworking/VimeoNetworking/Models/VIMSeason.h
```diff
@@ -17,4 +17,11 @@
 - (nullable VIMConnection *)connectionWithName:(nonnull NSString *)connectionName;
 - (nullable VIMInteraction *)interactionWithName:(nonnull NSString *)name;
 
+/**
+ Reports whether this season is protected by DRM.
+
+ @return A BOOL indicating whether the receiver is a DRM-protected season.
+ */
+- (BOOL)isDRMProtected;
```
ghking: Again maybe we can make this more descriptive. Since the content is DRM protected rather than the season object itself, maybe `containsDRMProtectedContent` or something

VimeoNetworking/VimeoNetworking/Models/VIMVideo.m
```diff
@@ -437,7 +437,16 @@ - (BOOL)isRatedMature
...
+    VIMInteraction *buyInteraction = [self interactionWithName:VIMInteractionNameBuy];
+    BOOL isBuyDRMProtected = buyInteraction.drm;
+    
+    VIMInteraction *rentInteraction = [self interactionWithName:VIMInteractionNameRent];
+    BOOL isRentDRMProtected = rentInteraction.drm;
+    
+    VIMInteraction *subscribeInteraction = [self interactionWithName:VIMInteractionNameSubscribe];
+    BOOL isSubscribeDRMProtected = subscribeInteraction.drm;
+    
+    return isBuyDRMProtected || isRentDRMProtected || isSubscribeDRMProtected;
```
ghking: Is this still necessary now that we've enabled the TV app on the API side to pull the fairplay files?

ghking:
DWR 👍

VimeoNetworking/VimeoNetworking/Models/VIMInteraction.h
jasonhawkins: I made some changes and tried to give this a more descriptive name. So now we can check for `someInteraction.isForDRMProtectedContent`.

VimeoNetworking/VimeoNetworking/Models/VIMVideo.m
jasonhawkins: I was worried that, because our app hadn't been enabled for DRM, the `drmFiles` array was nil even though we _did_ actually have enough information to determine whether the content was protected by DRM. 

Because the `drm` property isn't suppressed for apps that aren't DRM-enabled we can still get an accurate picture of what is and isn't covered by DRM. I agree it's not ideal though. But I like it more than looking at the `drmFiles` array, which doesn't seem to be as reliable.

VimeoNetworking/VimeoNetworking.xcodeproj/project.pbxproj
```diff
@@ -484,36 +473,53 @@
...
+				3C6F88771D998086006A24F4 /* VIMVideoProgressiveFile.h */,
+				3C6F88761D998086006A24F4 /* VIMVideoProgressiveFile.m */,
 				01FD38F71CC948F200326408 /* VIMVideoUtils.h */,
 				01FD38F81CC948F200326408 /* VIMVideoUtils.m */,
-				3C91DA321CCFEBF200DCF8BD /* VIMSoundtrack.h */,
-				3C91DA331CCFEBF200DCF8BD /* VIMSoundtrack.m */,
-				018483871CE0D9E2008305A4 /* PinCodeInfo.swift */,
-				3C6F88581D9972B3006A24F4 /* VIMProgrammedContent.swift */,
-				3CB75E211D22B8DD003826AE /* VIMThumbnailUploadTicket.h */,
-				3CB75E221D22B8DD003826AE /* VIMThumbnailUploadTicket.m */,
```
alfiehanssen: This is niiiiiiiiice!

VimeoNetworking/VimeoNetworking/Models/VIMSeason.m
```diff
@@ -85,4 +85,18 @@ - (void)parseInteractions
     self.interactions = interactions;
 }
 
+- (BOOL)containsDRMProtectedContent
+{
+    VIMInteraction *buyInteraction = [self interactionWithName:VIMInteractionNameBuy];
```
alfiehanssen: Will relying on `isForDRMProtectedContent` work in both the iOS and tvOS apps given that they are pointed at different API versions? 3.3 and 3.3.1

VimeoNetworking/VimeoNetworking/Models/VIMVideo.m
```diff
@@ -437,7 +437,16 @@ - (BOOL)isRatedMature
 
 - (BOOL)isDRMProtected
 {
-    return self.playRepresentation.drmFiles.fairPlayFile != nil;
+    VIMInteraction *buyInteraction = [self interactionWithName:VIMInteractionNameBuy];
```
alfiehanssen: This logic is in two classes, is there a way to consolidate it so it only appears once?

PR #86, opened 2016-12-19
Resolves build issues with VimeoNetworking

jasonhawkins:
#### Summary
Resolves all errors when building the iOS Example project and running unit tests against the iOS test target. Adds example project and unit test targets to support building against tvOS 9.0+. The structure of the project has changed as well, and is more in line with the file structure defined by CocoaPods.

#### Implementation Summary
- Added new files and paths to **.gitignore.** Also removed `-xcshareddata`
- Added **.swift-version** and specified Swift 2.3.
- Updated the **README**. More changes will be needed before we push this to CocoaPods.org and release it as 1.0 though.
- Updated **Podfile**
    - Changed `xcodeproj` to `project`.
    - Added `shared_pods` definition.
    - Added targets for `VimeoNetworkingExample-tvOS` and `VimeoNetworkingExample-tvOSTests`.
    - Added an action to set `SWIFT_VERSION` in `post_install`. Note that this is widely reported as not being necessary as of CocoaPods 1.0, but Circle CI was failing the builds without it.
- Updated **VimeoNetworking.podspec**:
    - Resolved warnings, fixed "LICENSE" file naming, and removed the Model subspec.
    - Updated the source and resources values to split source files into separate directories. 
- Fixed compiler warnings and errors for multiple files.
- Added a shared scheme for new `VimeoNetworkingExample-tvOS` and `VimeoNetworkingExample-tvOSTests` targets. These schemes build against `VimeoNetworking.framework`.

#### How to Test
- Run `pod install` and select the **VimeoNetworkingExample-iOS** scheme. 
- Build and run, note that the example project runs.
- Run unit tests, note that all tests should pass.
- Select the **VimeoNetworkingiOS** scheme.
- Build, note that the example project builds.
- Select the **VimeoNetworkingtvOS** scheme.
- Build, note that the example project builds.


ghking:
Looks siqqqqqqqqq

VimeoNetworking/Sources/ResponseCache.swift
```diff
@@ -150,7 +150,7 @@ final internal class ResponseCache
                 let fileManager = NSFileManager()
                 
                 let directoryURL = self.cachesDirectoryURL()
-                let fileURL = self.fileURLForKey(key: key)
+                _ = self.fileURLForKey(key: key)
```
jasonhawkins: It wasn't, no. And now that I look at this again I can see that just below here we have
```Swift
guard let directoryPath = directoryURL.path,
    let filePath = self.fileURLForKey(key: key)?.path
else
```
So I think maybe this was a mistake. I'm going to remove the line.

VimeoNetworking/Sources/VimeoSessionManager+Constructors.swift
```diff
@@ -132,17 +132,6 @@ public extension VimeoSessionManager
...
-        {
-            sessionConfiguration = NSURLSessionConfiguration.backgroundSessionConfigurationWithIdentifier(identifier)
-        }
-        else
-        {
-            sessionConfiguration = NSURLSessionConfiguration.backgroundSessionConfiguration(identifier)
-        }
-        
-        return sessionConfiguration
+        return NSURLSessionConfiguration.backgroundSessionConfigurationWithIdentifier(identifier)
```
jasonhawkins: Yeah, the deployment target for the pod is set for iOS 8+ and tvOS 9+. The deployment target for the example project is 9.3, but we could change that if we need to.

VimeoNetworking/Sources/Models/VIMVODItem.m
```diff
@@ -291,4 +291,4 @@ - (BOOL)canDoInteraction:(NSString *)interactionName
     return interaction && [interaction.uri length];
 }
 
-@end
\ No newline at end of file
+@end
```
jasonhawkins: squint

PR #88, opened 2017-01-05
VIM-4820: Added a new property fps to VIMVideoFile model object.

jenniferlim:
## Ticket

https://vimean.atlassian.net/browse/VIM-4820

## Ticket Summary

Cast video are sometimes choppy.

## Ticket Implementation

Chromecast does not support 2k/4k
The Chromecast processor can decode up to 720p at 60fps and 1080p at 30fps.
So we need to retrieve the fps value and make sure that the selected video to cast can be decoded by the chromecast processor.
 
## How to Test
Check that cast video works.

### DIFF
--- a/VimeoNetworking/VimeoNetworking/Models/VIMVideoFile.h
+++ b/VimeoNetworking/VimeoNetworking/Models/VIMVideoFile.h
@@ -39,6 +39,7 @@ extern NSString *const __nonnull VIMVideoFileQualityMobile;
 @property (nonatomic, strong, nullable) NSNumber *width;
 @property (nonatomic, strong, nullable) NSNumber *height;
 @property (nonatomic, strong, nullable) NSNumber *size;
+@property (nonatomic, strong, nullable) NSNumber *fps;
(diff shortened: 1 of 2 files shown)


nicolelehrer:
@jenniferlim this is a legacy class we don't use internally but it's great to keep it updated! looks great 👍 Done with Review

jasonhawkins:
What's the status of this PR? Are there iOS-side changes waiting on this one getting in?

jenniferlim:
@jasonhawkins - This only update `VIMVideoFile.h` model by adding a property `fps`. However, it's not use it internally according to @nicolelehrer in our app. So there is no iOS-side changes waiting or that to get in. We can either update VimeoNetworking or just close the PR. What do you guys think? cc @nicolelehrer

nicolelehrer:
@jenniferlim I agree with you, you can close this 👍  
Just be sure to update what commit you point to in your Vimeo-iOS branch
cc: @jasonhawkins

jenniferlim:
Alright, closing it.

PR #90, opened 2017-01-09
VIM-4824: Added Spatial model

jenniferlim:
## Ticket

https://vimean.atlassian.net/browse/VIM-4824

## Ticket Summary

Retrieve Spatial object for 360

## Implementation Summary

Added a new object in VimeoNetworking to retrieve the Spatial datas.
Updated the filter result in VideoStream to accept to retrieve the Spatial datas.

## How to Test
TBD

### DIFF
--- a/VimeoNetworking/VimeoNetworking/Models/Spatial.swift
+++ b/VimeoNetworking/VimeoNetworking/Models/Spatial.swift
@@ -27,6 +27,7 @@ public class Spatial: VIMModelObject
     
     // MARK: - Properties
     
+    // TODO: - Update to be Projection, Format types
     public dynamic var projection: String?
     public dynamic var stereoFormat: String?
     
--- a/VimeoNetworking/VimeoNetworking/Models/Spatial.swift
+++ b/VimeoNetworking/VimeoNetworking/Models/Spatial.swift
@@ -1,0 +1,39 @@
+//
+//  Spatial.swift
+//  Pods
+//
+//  Created by Lim, Jennifer on 1/6/17.
+//
+//
+//
+
+public class Spatial: VIMModelObject
+{
+    enum Projection: String
+    {
+        case Equirectangular = "equirectangular"
+        case Cylindrical = "cylindrical"
+        case Cubical = "cubical"
+        case Pyramid = "pyramid"
+        case Dome = "dome"
+    }
+    
+    enum Format: String
+    {
+        case Mono = "mono"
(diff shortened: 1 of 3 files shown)


VimeoNetworking/VimeoNetworking/Models/Spatial.swift
```diff
@@ -0,0 +1,40 @@
...
+//  Created by Lim, Jennifer on 1/6/17.
+//
+//
+//
+
+public class Spatial: VIMModelObject
+{
+    enum Projection: String
+    {
+        case Equirectangular = "equirectangular"
```
jasonhawkins: I'm torn as to whether these strings should live here or not. I think maybe it's okay? I'm not sure putting them in a Constants struct would be much of an improvement so I guess just leave it.

VimeoNetworking/VimeoNetworking/Models/Spatial.swift
```diff
@@ -0,0 +1,40 @@
...
+    {
+        case Mono = "mono"
+        case LeftRight = "left-right"
+        case TopBottom = "top-bottom"
+    }
+    
+    // MARK: - Properties
+    
+    // TODO: - Update to be Projection, Format types
+    public dynamic var projection: String?
```
jasonhawkins: These probably don't need to be re-declared as public since the class is already public.

VimeoNetworking/VimeoNetworking/Models/Spatial.swift
```diff
@@ -0,0 +1,40 @@
...
+    public dynamic var stereoFormat: String?
```
jasonhawkins: I wonder if it's worth making `typealias`es for these strings. So that the Format enum is also of type `FormatType`, which would be a typealias for String. And then something similar for projection.

jasonhawkins:
Hey Jennifer! Sorry for the delay on this (and the related ones). Can you pull in the latest from `develop` and make sure your new Spatial.swift model is located in the `VimeoNetworking/VimeoNetworking/Sources/Models` directory? You'll then have to run a `pod install` to get the file to show up in the Pods project, but let me know if you need any help or have questions about it. 

Since **Spatial** is going to be a public class we should also add some documentation in there, for anything that's going to be public.

DWR for now. Let me know when the changes are up and I'll check it out! 🐳

jenniferlim:
please see: https://github.com/vimeo/VimeoNetworking/pull/95

PR #92, opened 2017-01-09
Set Use Legacy Swift Verison to YES.

jasonhawkins:
We had mismatched Swift versions for some of the targets. This resolves that.

`pod install` should work again.


jasonhawkins:
Thanks, Mike!

PR #95, opened 2017-01-19
VIM-4824: Three sixty model

jenniferlim:
## Ticket

https://vimean.atlassian.net/browse/VIM-4824

## Ticket Summary

Retrieve Spatial object for 360

## Implementation Summary

Added a new object in VimeoNetworking to retrieve the Spatial datas.
Updated the filter result in VideoStream to accept to retrieve the Spatial datas.

## How to Test
TBD


VimeoNetworking/Sources/Models/Spatial.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+//
+//
+
+/// Spatial stores all information related to threesixty video
+public class Spatial: VIMModelObject
+{
+    
+    /// Represents the projection. Value returned by the server can be: "equirectangular", "cylindrical", "cubical", "pyramid", "dome".
+    dynamic var projection: String?
```
jasonhawkins: `dynamic` so we can KVO these properties from Objective-C?

VimeoNetworking/Sources/Models/Spatial.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    /// Represents the format. Value returned by the server can be: "mono", left-right", "top-bottom"
```
jasonhawkins: Awesome, hooray for documentaaaationnnnn

VimeoNetworking/Sources/Models/Spatial.swift
jenniferlim: that was the idea, but finally i don't think we need to kvo this. I'll update the PR.

jasonhawkins: I guess if we ever need to take advantage of KVO for these then we can change it back. In the meantime we'll enjoy that sweet, sweet minuscule and imperceptible performance improvement by not making it dynamic! Woooo 🏃🏻‍♀️💨

jenniferlim:
@jasonhawkins - Thanks for your feedback, I updated my PR. Let me know if we are good to 📫 ?

jasonhawkins:
I think you're good to go here, thanks Jennifer 👍🏼

jenniferlim:
Awesome ⭐️

PR #108, opened 2017-02-08
VIM-4764: Adding 360 DRM Error

ghking:
See https://github.vimeows.com/MobileApps/Vimeo-iOS/pull/1165

### DIFF
--- a/VimeoNetworking/Sources/Models/VIMVideo.m
+++ b/VimeoNetworking/Sources/Models/VIMVideo.m
@@ -436,6 +436,15 @@ NSString *VIMContentRating_Safe = @"safe";
 
 - (BOOL)isDRMProtected
 {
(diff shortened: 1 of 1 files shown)


VimeoNetworking/Sources/Models/VIMVideo.m
```diff
@@ -436,6 +436,15 @@ - (BOOL)isRatedMature
 
 - (BOOL)isDRMProtected
 {
+    if (self.playRepresentation.drmFiles.fairPlayFile)
```
nicolelehrer: Should we swap consideration of these cases? I remember for VOD, we would sometimes have access to files we shouldn't so we had to look at the play status to determine if it's allowed to be played back. I was wondering if there's a similar edge case here, where checking the stricter of the 2 cases first is worth it.

PR #117, opened 2017-02-28
CI-132 | Add Pods to git

wilrnh:
This PR
- removes `Pods/` from .gitignore
- removes `pod install` from Fastlane
- adds `Pods/` to git

To test:
- clean checkout should build/test successfully without `pod install`


MathiasGr:
@mikew-personal @jasonhawkins @wilrnh This change makes us distribute the pods through our own Repository, do we have legal rights to do so?

wilrnh:
Good question, and I just checked a little into this:

It looks like the only Pod we use atm is AFNetworking, which has an MIT license, so for now I think we're aok. But the question remains if we can sustain this setup if we ever need a more restricted library :(

PR #119, opened 2017-03-07
AF-18: Remove a specific cached file

nicolelehrer:
#### Ticket
[AF-18](https://vimean.atlassian.net/browse/AF-18)

#### Ticket Summary
We were seeing `"could not remove disk cache for \(key)"` printed in the console. 

In order to access a specific response in the cache, we use a cache key on the request object. This `cacheKey` is defined in `Request+Cache`. When we get a response back, we add it to the cache using `setResponseDictionary` for a given cache key. For our disk caching we write this response to disk. When we need to update what's on disk with fresh data, we call `removeResponseDictionary` and use a cache key to find the file we want to delete. 

This ticket originally called for resolving the difference between how we first generate a cache key and how we then re-generate that key to find the corresponding specific cached file. It was found that this process is in fact ok - we are generating the key in the same way, and use the same key to both store cached files and find them later. 

- This was validated by comparing the value for `cacheKeyForCurrentStream` (which we generate using an empty request object whenever we need to know what the cache key is for a particular stream) to the `cacheKey` property generated on the original request object. Also we could validate these keys matched by confirming a successful removal of a cached file found using this method. 
 
The reason why we were seeing `"could not remove disk cache for \(key)"` printed in the console was because we were trying to delete a cache file that had already been deleted.

- This was confirmed by checking a few use cases that use toggle actions: updating category objects post follow / unfollow. The cached file did exist at one point, was deleted when we updated one of the categories to be followed, and then when we tried to delete it again by updating another category in the list, the error said the file doesn't exist. You can also do this by liking a video in a list. 

Several stream datasources are also listening to notifications that result in invalidating their respective cached files following a toggleAction. So these streams invalidate their respective caches - and then if asked to do it again before they can repopulate the cache, it's possible to see multiple instances of the above mentioned log in the console.  

Note - When invalidating the cache, we create a fake request object so we can derive the cache key so that any stream's data source listening for updates can invalidate its own associated cache - we do this inside `RemoteSectionDataSource` which has the stream uri and parameters handy and can therefore be used to generate the cache key for its corresponding cached data source. While this isn't ideal, we currently cache response data at the resolution of the full stream response - so we need listeners to these update notifications to be able to delete their corresponding cache if needed. We can't broadcast the specific cache key of the file to delete since we need stream data sources to delete their own content based on their own cache key, generated using the request url and parameters. 

#### Implementation Summary

- only assert if `filePath` is nil
- if that succeeds, check to see if the cache file exists at that path. If it doesn't exist, then exit the method. 
- Otherwise we're safe to try to remove the cached file

#### How to Test
- Update to the PR branch listed in Vimeo-iOS PR https://github.vimeows.com/MobileApps/Vimeo-iOS/pull/1248
- Inside `ResponseDiskCache` - set a break point in the following places
- inside `setResponseDictionary` to inspect line `let success = fileManager.createFileAtPath(filePath, contents: data, attributes: nil)` to make sure a file for a given cache key was written to
- inside `removeResponseDictionary` to inspect the do-catch statement - you can verify that if the `try fileManager.removeItemAtPath(filePath)` is being called without the catch triggered, you have removed the file with success
- inside `removeResponseDictionary` to inspect the line `guard fileManager.fileExistsAtPath(filePath) == true`, you can verify that a file's not there
- the pattern we expect is, you can successfully remove a cached file once by modifying an object (liking a video on a stream) but if you try to do it again without refreshing the stream, the cached file won't be there
- keep in mind this can get tricky - several things are updated when we use a toggle action so these methods can be called multiple times, so make sure you're consistent with respect to which cached file you're inspecting.


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/62013f321b5db60db6ca91c2bacd4dec181f6d36/VimeoNetworking/Sources/VimeoClient.swift#L328">VimeoNetworking/Sources/VimeoClient.swift#L328</a></strong>: cast from ‘Response&lt;VIMNullResponse&gt;’ to unrelated type ‘Response&lt;ModelType&gt;’ always fails  <br /><code>
</code></td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/62013f321b5db60db6ca91c2bacd4dec181f6d36/VimeoNetworking/Sources/Models/VIMVideoPreference.m#L41">VimeoNetworking/Sources/Models/VIMVideoPreference.m#L41</a></strong>: incompatible pointer types assigning to ‘NSString * _Nullable’ from ‘VIMPrivacy * _Nullable’ [-Wincompatible-pointer-types]  <br /><code>
        privacy.view = self.privacy;
</code></td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 22 tests, with 0 failures (0 unexpected) in 0.330 (0.377) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/180)

</p>

PR #121, opened 2017-03-13
Swift 3 Conversion, Part 1

jasonhawkins:
- Includes the latest from the `develop` branch.
- Framework and sample project build, but otherwise has not been tested.
- Large sections of `result`-based code has been commented out of the sample project because of unresolved compiler errors.


ghking:
@jasonhawkins Any way to separate the Pods stuff from the actual conversion?

PR #129, opened 2017-03-31
VimeoKit Swift 3 Conversion

ghking:
See https://github.vimeows.com/MobileApps/VimeoKit/pull/163.

- Since `NSNotification` is now `Notification`, I had to fix some naming conflicts since VimeoNetworking also contains a `Notification` enum. I changed this to `NetworkingNotification`.
- Handled some cases where Any should be used rather than AnyObject.


jasonhawkins:
Looks like this need a merge from `swift-3-part-1` but otherwise I think you're good to go.

PR #133, opened 2017-04-17
TVOS-732: Cachekey

jenniferlim:
#### Ticket

[TVOS-732](https://vimean.atlassian.net/browse/TVOS-732)

#### Pull Request Checklist

#### Issue Summary

- When requesting a paginated result (!= firstPage), the cacheKey generated was wrong. That results by retrieving this issue:

`ResponseDiskCache could not store object`

#### Implementation Summary

- Updated the way we generate the cacheKey in `Request+Cache.swift`

- What happen before is that for the first request results: the `pathURI` didn't contain any `queryItems`: So the cacheKey generated for the first page was always correct. Although, for a request paginated, the `pathURI` contained the `queryItems`. Because, we were concatenated `cached` with the `pathURI`: We got a super long cacheKey.

The fix was to:
1 - create a `NSURL` object with `pathURI`. So that we can only retrieve the real `path` without the filters.

2 - Instead of getting a hashValue for the `parameters`, get the hashValue of the `pathURI` which will be always unique.

#### Reviewer Tips

- In `ResponseCache.swift` file, that method `private func setResponseDictionary(responseDictionary: VimeoClient.ResponseDictionary, forKey key: String)`:
-> Add some logs:
```
do
                {
                    if !fileManager.fileExistsAtPath(directoryPath)
                    {
                        try fileManager.createDirectoryAtPath(directoryPath, withIntermediateDirectories: true, attributes: nil)
                    }
                    
                    let success = fileManager.createFileAtPath(filePath, contents: data, attributes: nil)
                    
                    if !success
                    {
                        print("ResponseDiskCache could not store object\(filePath)")
                    }
                    else
                    {
                        print("yeayyyy\(filePath)")
                    }
                }
```

You'll notice that before the changes, requesting a `nextPage` will always reach the error: `ResponseDiskCache could not store object`.

#### How to Test

- Go to explore, and scroll to request another page. See that in the console, there is no error logs.
- Make sure, retrieving cache still works on both platform.

### DIFF
--- a/VimeoNetworking/Sources/Dictionary+Extension.swift
+++ b/VimeoNetworking/Sources/Dictionary+Extension.swift
@@ -34,4 +34,17 @@ extension Dictionary
             self[key] = value
         }
     }
+
+    // `toQueryString` transforms a dictionary to a queryString
+    public var toQueryString: String {
+        
+        var output: String = ""
+        
+        for (_, dictionary) in self.enumerated()
+        {
+            output += "&\(dictionary.key)=\(dictionary.value)"
+        }
+
+        return output
+    }
 }
--- a/VimeoNetworking/Sources/Request+Cache.swift
+++ b/VimeoNetworking/Sources/Request+Cache.swift
@@ -32,12 +32,10 @@ public extension Request
     /// Generates a unique cache key for a request, taking into account endpoint and parameters
     var cacheKey: String
     {
-        var cacheKey = "cached" + self.path
+        let url = NSURL(string: self.path)
+        let urlPath = url?.path ?? ""
         
-        if let description = (self.parameters as? CustomStringConvertible)?.description
-        {
-            cacheKey = cacheKey + "." + String(description.hashValue)
-        }
+        var cacheKey = "cached" + urlPath + "." + String(self.path.hashValue)
         
         cacheKey = cacheKey.replacingOccurrences(of: "/", with: ".")
         
--- a/VimeoNetworking.xcworkspace/xcshareddata/VimeoNetworking.xcscmblueprint
+++ b/VimeoNetworking.xcworkspace/xcshareddata/VimeoNetworking.xcscmblueprint
@@ -1,0 +1,30 @@
+{
+  "DVTSourceControlWorkspaceBlueprintPrimaryRemoteRepositoryKey" : "C3D71035A96924D900556ABAF7AF12CFE47E2A44",
+  "DVTSourceControlWorkspaceBlueprintWorkingCopyRepositoryLocationsKey" : {
+
+  },
+  "DVTSourceControlWorkspaceBlueprintWorkingCopyStatesKey" : {
+    "C3D71035A96924D900556ABAF7AF12CFE47E2A44" : 9223372036854775807,
+    "46A27ADFCE516C0BE2A99EB615FE0DF976CA40FC" : 9223372036854775807
+  },
+  "DVTSourceControlWorkspaceBlueprintIdentifierKey" : "0658EF35-AF12-4359-8927-015EAC095B38",
+  "DVTSourceControlWorkspaceBlueprintWorkingCopyPathsKey" : {
+    "C3D71035A96924D900556ABAF7AF12CFE47E2A44" : "VimeoNetworking\/",
+    "46A27ADFCE516C0BE2A99EB615FE0DF976CA40FC" : ".."
+  },
+  "DVTSourceControlWorkspaceBlueprintNameKey" : "VimeoNetworking",
+  "DVTSourceControlWorkspaceBlueprintVersion" : 204,
+  "DVTSourceControlWorkspaceBlueprintRelativePathToProjectKey" : "VimeoNetworking.xcworkspace",
+  "DVTSourceControlWorkspaceBlueprintRemoteRepositoriesKey" : [
+    {
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryURLKey" : "https:\/\/github.vimeows.com\/MobileApps\/Vimeo-tvOS.git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositorySystemKey" : "com.apple.dt.Xcode.sourcecontrol.Git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryIdentifierKey" : "46A27ADFCE516C0BE2A99EB615FE0DF976CA40FC"
+    },
+    {
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryURLKey" : "github.com:vimeo\/VimeoNetworking.git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositorySystemKey" : "com.apple.dt.Xcode.sourcecontrol.Git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryIdentifierKey" : "C3D71035A96924D900556ABAF7AF12CFE47E2A44"
+    }
+  ]
+}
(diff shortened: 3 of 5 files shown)


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/d4cb07d94257057b1818a471470e805988f973f8/VimeoNetworking/Sources/Request+Notifications.swift#L73">VimeoNetworking/Sources/Request+Notifications.swift#L73</a></strong>: result of call to ‘map’ is unused  <br /><code>
        notifications.map { (notification: VIMNotification) -&gt; Void in
</code></td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/d4cb07d94257057b1818a471470e805988f973f8/VimeoNetworking/Sources/VimeoClient.swift#L328">VimeoNetworking/Sources/VimeoClient.swift#L328</a></strong>: cast from ‘Response&lt;VIMNullResponse&gt;’ to unrelated type ‘Response&lt;ModelType&gt;’ always fails  <br /><code>
</code></td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/d4cb07d94257057b1818a471470e805988f973f8/VimeoNetworking/Sources/Models/VIMVideoPreference.m#L41">VimeoNetworking/Sources/Models/VIMVideoPreference.m#L41</a></strong>: incompatible pointer types assigning to ‘NSString * _Nullable’ from ‘VIMPrivacy * _Nullable’ [-Wincompatible-pointer-types]  <br /><code>
        privacy.view = self.privacy;
</code></td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/d4cb07d94257057b1818a471470e805988f973f8/VimeoNetworking/Sources/Models/VIMVODItem.m#L285">VimeoNetworking/Sources/Models/VIMVODItem.m#L285</a></strong>: incompatible pointer types returning ‘VIMConnection * _Nullable’ from a function with result type ‘VIMVODConnection *’ [-Wincompatible-pointer-types]  <br /><code>
    return [self connectionWithName:VIMConnectionNameVideos];
</code></td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 24 tests, with 0 failures (0 unexpected) in 0.331 (0.343) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/269)

</p>

VimeoNetworking/Sources/Request+Cache.swift
```diff
@@ -32,12 +32,10 @@ public extension Request
     /// Generates a unique cache key for a request, taking into account endpoint and parameters
     var cacheKey: String
     {
-        var cacheKey = "cached" + self.path
+        let url = NSURL(string: self.path)
+        let urlPath = url?.path ?? ""
```
ghking: Aren't `self.path` and `urlPath` the same thing? Do we even need to construct an NSURL?

VimeoNetworking/Sources/Request+Cache.swift
```diff
@@ -32,12 +32,10 @@ public extension Request
...
         
-        if let parameters = self.parameters
-        {
-            cacheKey = cacheKey + "." + String(parameters.description.hashValue)
-        }
+        var cacheKey = "cached" + urlPath + "." + String(self.path.hashValue)
```
ghking: OK so I see we're ignoring the parameters. Is that because we're assuming anything we cache will be a `GET`, and therefore shouldn't theoretically have a request body? I think this makes sense, but maybe we should throw some guards with asserts up top to make sure that `self.method == GET` and `self.parameters == nil`.

VimeoNetworking/Sources/Request+Cache.swift
jenniferlim: @ghking - No, they are different and it's confusing because:
- `self.path` should only contain the `path` without the `URI`. However, because `self.path` is a `string` it contains the `path+uri`.
- `urlPath` is needed to extract the real path (= without the `URI`).

VimeoNetworking/Sources/Request+Cache.swift
jenniferlim: 1) What we tried to achieve in the past was to generate a unique cached key based on the parameters. However, the parameter doesn't change when you do a paginated request.

2) What we want to do instead: based on the `self.path` that is equal to the `uri` generate a uniqueKey. That should be enough because the uri contains the `real path + the queryItems (= variable when you paginate)`.

ghking:
Why are we trying to store second page requests on tvOS? Are we doing that anymore?

Secondly, it looks like our first page requests have a `parameters` body, generated by `configuredRequestParameters()` in `RemoteSectionDataSource`. Since this is a GET request, this goes against HTTP standards.  I know this was existing logic (that I probably wrote), but now that I'm looking into it I have no idea why we're doing that. For first page request, I think any data should be appended to the URL as a query rather than as a 'parameters' body. Then, to generate the `cacheKey` for any GET request, we'd use pretty much the same logic that's currently in place except we'd append a `hashValue` of the query rather than a `hashValue` of the `parameters` body.

Basically I think the weirdness is coming from the fact that in some cases the filters are in the body (fist page requests) and sometimes they're in the URL (all other page requests). Maybe we should just convert `configuredRequestParameters()` to something like `configuredRequestQueryString()`? Maybe that would make more sense and would unify the format of our requests. Does that make sense?

jenniferlim:
@ghking - As suggested, i implemented `configuredRequestQueryString` and added an extension to `Dictionary` so that we can get a `queryString` from a `Dictionary`.
It's ready for another look, thanks.

VimeoNetworking/Sources/Dictionary+Extension.swift
```diff
@@ -34,4 +34,17 @@ extension Dictionary
...
+        
+        var output: String = ""
+        
+        for (_, dictionary) in self.enumerated()
+        {
+            output += "&\(dictionary.key)=\(dictionary.value)"
+        }
+
+        return output
+    }
```
ghking: For an example, throw this code in a Playground:

```
import Foundation

let queryDictionary = ["foo": "bar", "baz": "blah"]

var urlComponents = URLComponents()
urlComponents.path = "/users/gavin"
urlComponents.queryItems = queryDictionary.map({ (key, value) in URLQueryItem(name: key, value: value) })

print(urlComponents.string!)
```

VimeoNetworking/Sources/Request+Cache.swift
```diff
@@ -32,12 +32,10 @@ public extension Request
...
     {
-        var cacheKey = "cached" + self.path
+        let url = NSURL(string: self.path)
+        let urlPath = url?.path ?? ""
         
-        if let description = (self.parameters as? CustomStringConvertible)?.description
-        {
-            cacheKey = cacheKey + "." + String(description.hashValue)
-        }
+        var cacheKey = "cached" + urlPath + "." + String(self.path.hashValue)
```
ghking: Should guard and assert (and return nil?) if `url.path` is null rather than doing `let urlPath = url?.path ?? "`? I think the latter case would indicate that something bad happened.

jenniferlim:
See https://github.com/vimeo/VimeoNetworking/pull/146

PR #134, opened 2017-04-17
removing a guard statememnt and replacing it with default values for …

mikew-personal:
#### Ticket

N/A

#### Pull Request Checklist

#### Issue Summary

Some API endpoints don't always return all of the expected paging information.  For instance, the /me/feed endpoint can't return the total number of items so there is no "total" key in the JSON response.  Currently these requests are causing the app to assert because we are expecting that the data will always be there.  This PR removes a guard statement looking for the data and replaces it with default values for when the data is missing.

#### Implementation Summary

- VimeoClient - removed the guard statement checking for totalCount, currentPage, and itemsPerPage and replaced it with a default value of 0 if any of those keys are missing from the JSON response.

#### Reviewer Tips

n/a

#### How to Test

1. Make sure the Feed is accessible in the tvOS app.

### DIFF
--- a/VimeoNetworking/Sources/VimeoClient.swift
+++ b/VimeoNetworking/Sources/VimeoClient.swift
@@ -352,14 +352,9 @@ final public class VimeoClient
             
             if let pagingDictionary = responseDictionary[Constants.PagingKey] as? ResponseDictionary
             {
-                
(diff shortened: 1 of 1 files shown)


jasonhawkins:
I've included these changes in the main [Swift 3 PR](https://github.com/vimeo/VimeoNetworking/pull/135). We should be okay to close this now.

PR #143, opened 2017-05-15
[CRI-108] Singleton for VimeoClient

fyell:
#### Ticket

[CRI-108](https://vimean.atlassian.net/browse/CRI-108)

- *Note: this is required for Vimeo staff only. If not applicable to your PR, use "N/A".*

#### Pull Request Checklist

#### Issue Summary

Since internal frameworks rely on a previously extracted singleton instance, we needed to add a way for clients to configure the singleton instance

#### Implementation Summary

- Made AppConfiguration and VimeoSessionManager on VimeoClient optional
- Add static methods for singleton and singleton configuration

### DIFF
--- a/VimeoNetworking/Sources/AuthenticationController.swift
+++ b/VimeoNetworking/Sources/AuthenticationController.swift
@@ -82,13 +82,19 @@ final public class AuthenticationController
      
      - returns: a new `AuthenticationController`
      */
-    public init(client: VimeoClient)
+    public init?(client: VimeoClient)
     {
-        self.configuration = client.configuration
-        self.client = client
-        self.accountStore = AccountStore(configuration: client.configuration)
+        guard let configuration = client.configuration else
+        {
+            assertionFailure("VimeoClient argument cannot contain nil configuration")
+            return nil
+        }
         
-        self.authenticatorClient = VimeoClient(appConfiguration: client.configuration)
+        self.configuration = configuration
+        self.client = client
+        self.accountStore = AccountStore(configuration: configuration)
+        
+        self.authenticatorClient = VimeoClient(appConfiguration: configuration)
     }
     
     // MARK: - Public Saved Accounts
--- a/VimeoNetworking/Sources/AuthenticationController.swift
+++ b/VimeoNetworking/Sources/AuthenticationController.swift
@@ -82,15 +82,9 @@ final public class AuthenticationController
      
      - returns: a new `AuthenticationController`
      */
-    public init?(client: VimeoClient)
+    public init(client: VimeoClient, appConfiguration: AppConfiguration)
     {
-        guard let configuration = client.configuration else
-        {
-            assertionFailure("VimeoClient argument cannot contain nil configuration")
-            return nil
-        }
-        
-        self.configuration = configuration
+        self.configuration = appConfiguration
         self.client = client
         self.accountStore = AccountStore(configuration: configuration)
         
--- a/VimeoNetworking/Sources/VimeoClient.swift
+++ b/VimeoNetworking/Sources/VimeoClient.swift
@@ -87,7 +87,7 @@ final public class VimeoClient
     // MARK: -
     
         /// Session manager handles the http session data tasks and request/response serialization
-    private let sessionManager: VimeoSessionManager
+    fileprivate var sessionManager: VimeoSessionManager? = nil
     
         /// response cache handles all memory and disk caching of response dictionaries
     private let responseCache = ResponseCache()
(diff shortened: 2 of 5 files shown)


fyell:
build this please

VimeoNetworking/Sources/VimeoClient.swift
```diff
@@ -482,3 +486,34 @@ final public class VimeoClient
...
+        guard let _ = self._sharedClient.configuration,
+            let _ = self._sharedClient.sessionManager else
+        {
+            assertionFailure("VimeoClient.sharedClient must be configured before accessing")
+            return self._sharedClient
+        }
+        
+        return self._sharedClient
+    }
+    private static let _sharedClient = VimeoClient(appConfiguration: nil, sessionManager: nil)
```
nicolelehrer: It would be great to consolidate config with instantiation so we don't have to make these params optional or guard against folks using the singleton prior to config - 

Can we instead  pass in these params into an init method and configure the properties on the singleton before returning it? We could define how to configure the params by implementing a static method via an extension on `VimeoClient`

Could something like this work: 

```swift
static let sharedInstance : VimeoClient = {
        let instance = VimeoClient(config: VimeoClient.getAppConfig())
        return instance
}()

init( config : AppConfigType) {
        self.config = config
}

extension VimeoClient
{
    static func getAppConfig() -> AppConfigType
    {
        return appConfigInstance
    }
}
```

fyell:
Ahh. I think this was the syntax I was looking for. So does that mean that
block is only called once for sharedInstance?

fyell:
This approach also makes the framework assume that the client is
implementing an extension with a method called getAppConfig(). This means
that the framework won't compile without the client.

nicolelehrer:
that's a good point @fyell.

VimeoNetworking/Sources/AuthenticationController.swift
```diff
@@ -82,13 +82,13 @@ final public class AuthenticationController
...
+    public init(client: VimeoClient, appConfiguration: AppConfiguration)
     {
-        self.configuration = client.configuration
+        self.configuration = appConfiguration
         self.client = client
-        self.accountStore = AccountStore(configuration: client.configuration)
+        self.accountStore = AccountStore(configuration: configuration)
         
-        self.authenticatorClient = VimeoClient(appConfiguration: client.configuration)
+        self.authenticatorClient = VimeoClient(appConfiguration: configuration)
```
nicolelehrer: Ah ok this was pointed out by @mikew-personal - why would we not just pass the same client into the authenticator. Seems like we should be able to do `self.authenticatorClient = client` here. I know this was already in place before your changes @fyell but any thoughts on this?

fyell:
I haven't looked into the reasoning for creating another instance of the
client. But I feel we can address this outside of the scope of this PR just
so we can isolate the change if an issue occurs.

VimeoNetworking/Sources/AuthenticationController.swift
nicolelehrer: @mikew-personal @fyell here's the answer         
/// We need to use a separate client to make the actual auth requests, to ensure it's unauthenticated
    private let authenticatorClient: VimeoClient

nicolelehrer:
👍  thanks @fyell I resolved my comment with a followup

fyell:
build this please

cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><strong><a href="https://github.com/vimeo/VimeoNetworking/blob/09959936ea4cdd0f4b2c69e42268126349c6a21d/VimeoNetworking/Sources/VimeoClient.swift#L330">VimeoNetworking/Sources/VimeoClient.swift#L330</a></strong>: cast from ‘Response&lt;VIMNullResponse&gt;’ to unrelated type ‘Response&lt;ModelType&gt;’ always fails  <br /><code>
</code></td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 22 tests, with 0 failures (0 unexpected) in 0.360 (0.374) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/330)

</p>

PR #150, opened 2017-05-30
Merging Release 2.0.0

mikew-personal:
#### Issue Summary

- Merging in the release 2.0.0 branch.  This branch includes a number of new classes + bug fixes, as well as support for Swift 3.


.swift-version
```diff
@@ -1 +1 @@
-2.3
+3.1
```
ghking: Shouldn't this be `3.0`?

ghking: cc @jasonhawkins

jasonhawkins: It should have been, yeah. I'm kind of surprised CocoaPods didn't shout about it during pod install. I think if it didn't cause a problem for the release builds then maybe we should leave it, unless you guy have strong feelings about it.

PR #153, opened 2017-06-12
AF-83: Modifying User Agent

ghking:
#### Ticket

https://vimean.atlassian.net/browse/AF-83

#### Pull Request Checklist

#### Issue Summary

Add the framework string to the user agent: https://github.vimeows.com/Vimeo-API/api/issues/1141#issuecomment-143572

#### Implementation Summary

Appending the framework name and version.

#### How to Test

Set a breakpoint to make sure the user agent looks like it should (described in the ticket).

### DIFF
--- a/VimeoNetworking/Sources/VimeoRequestSerializer.swift
+++ b/VimeoNetworking/Sources/VimeoRequestSerializer.swift
@@ -37,6 +37,7 @@ final public class VimeoRequestSerializer: AFJSONRequestSerializer
     {
         static let AcceptHeaderKey = "Accept"
         static let AuthorizationHeaderKey = "Authorization"
+        static let UserAgentKey = "User-Agent"
     }
     
     public typealias AccessTokenProvider = (Void) -> String?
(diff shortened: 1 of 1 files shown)


VimeoNetworking/Sources/VimeoRequestSerializer.swift
```diff
@@ -169,4 +180,33 @@ final public class VimeoRequestSerializer: AFJSONRequestSerializer
         
         return request
     }
+    
+    private func requestModifyingUserAgentHeader(fromRequest request: URLRequest) -> URLRequest
+    {
+        guard let frameworkVersion = Bundle(for: type(of: self)).infoDictionary?["CFBundleShortVersionString"] as? String else
```
jasonhawkins: Curious if we have to use `type(of: self)` here or if it'll allow us to just do `Bundle(for: VimeoRequestSerializer)`. Would that work?

ghking: It would, but I'd rather not hardcode the class name if we don't have to.

PR #176, opened 2017-08-15
Merge release/3.0.1 re-integration branch

mikew-personal:
#### Pull Request Checklist

#### Issue Summary

- Re-integrates bug fixes that were made on the release/3.0.1 branch.

### DIFF
--- a/VimeoNetworking/Sources/Models/VIMQuantityQuota.m
+++ b/VimeoNetworking/Sources/Models/VIMQuantityQuota.m
@@ -39,8 +39,8 @@
 
 - (void)didFinishMapping
 {
-    self.canUploadHd = [self.hd boolValue];
-    self.canUploadSd = [self.sd boolValue];
+    self.canUploadHd = [self.hd respondsToSelector:@selector(boolValue)] ? [self.hd boolValue] : NO;
+    self.canUploadSd = [self.sd respondsToSelector:@selector(boolValue)] ? [self.sd boolValue] : NO;
 }
 
 @end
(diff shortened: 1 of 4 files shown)


VimeoNetworking/Sources/Models/VIMQuantityQuota.m
```diff
@@ -39,8 +39,8 @@ @implementation VIMQuantityQuota
 
 - (void)didFinishMapping
 {
-    self.canUploadHd = [self.hd boolValue];
-    self.canUploadSd = [self.sd boolValue];
+    self.canUploadHd = [self.hd respondsToSelector:@selector(boolValue)] ? [self.hd boolValue] : NO;
+    self.canUploadSd = [self.sd respondsToSelector:@selector(boolValue)] ? [self.sd boolValue] : NO;
```
ghking: Why would this not respond to bool value?

jasonhawkins: I think we ran into a situation where we weren't getting expected values from the API for`NSNumber *hd` and ` NSNumber *sd`. I don't know what it was actually sending but whatever it was it couldn't handle being asked about its `boolValue` which was causing the 💥

mikew-personal: we don't know what the exact cause is yet, but the API was essentially sending back empty objects as the value instead of bool, so the object mapper converted those objects into NSDictionary objects which don't respond to boolValue.  The closest we've been to seeing this in house was a week or two ago when one of the VMs was broken and sending out bad data for all requests.

PR #210, opened 2018-02-21
AF-211: Adds compatibility with Swift 4

jasonhawkins:
### Ticket
[AF-211](https://vimean.atlassian.net/browse/AF-211)

### Ticket Summary
Add support and compatibility with Swift 4.

### Implementation Summary
- Added `@objc` and `@nonobjc` attributes to Swift source files to designate Objective-C accessibility.
- Removed support for Swift 3 style Objective-C interface inference.
- Added an `account` property on **VIMLiveChatUser** (internal use only :wink:) to allow access to account type from Objective-C.
- Addressed a deprecation warning in **Request** regarding use of the `characters` property.
- Updated documentation.

### How to Test
- Run unit tests.
- Try integrating in a mixed-source project. Ensure expected APIs are available from Objective-C.


VimeoNetworking/Sources/ResponseCache.swift
```diff
@@ -40,7 +40,7 @@ private protocol Cache
 }
 
 /// Response cache handles the storage of JSON response dictionaries indexed by their associated `Request`.  It contains both memory and disk caching functionality
-final internal class ResponseCache
+final public class ResponseCache
```
jenniferlim: is it intentional to make this guy public?

jasonhawkins: Yeaaah, since it was internally accessible to the module, but not accessible to the test targets which live outside of that.

VimeoNetworking/Sources/Constants.swift
```diff
@@ -30,4 +30,4 @@ import Foundation
 public let VimeoBaseURL = URL(string: "https://api.vimeo.com")!
 
  /// Default API version to use for requests
-internal let VimeoDefaultAPIVersionString = "3.2"
+public let VimeoDefaultAPIVersionString = "3.2"
```
ghking: why public?

VimeoNetworking/Sources/ExceptionCatcher+Swift.swift
```diff
@@ -35,7 +35,7 @@ class ExceptionCatcher: ObjC_ExceptionCatcher
      
      - throws: an error containing any thrown exception information
      */
-    @nonobjc internal static func doUnsafe(unsafeBlock: @escaping ((Void) -> Void)) throws
+    @nonobjc public static func doUnsafe(unsafeBlock: @escaping (() -> Void)) throws
```
ghking: Why public for this class and method?

VimeoNetworking/Sources/Models/Spatial.swift
```diff
@@ -10,9 +10,9 @@
 /// Spatial stores all information related to threesixty video
 public class Spatial: VIMModelObject
 {
-    public static let StereoFormatMono = "mono"
-    public static let StereoFormatLeftRight = "left-right"
-    public static let StereoFormatTopBottom = "top-bottom"
+    @objc public static let StereoFormatMono = "mono"
+    @objc public static let StereoFormatLeftRight = "left-right"
+    @objc public static let StereoFormatTopBottom = "top-bottom"
```
ghking: 👍

VimeoNetworking/Sources/Models/VIMLiveChatUser.swift
```diff
@@ -47,12 +47,36 @@ public enum AccountType: String
...
+    
+    /// Objecive-C accessible constant for the "Vimeo Plus" tier.
+    @objc public static let AccountTypePlus = "plus"
+    
+    /// Objecive-C accessible constant for the "Vimeo PRO" tier.
+    @objc public static let AccountTypePro = "pro"
+    
+    /// Objecive-C accessible constant for the "Custom Live" tier.
+    @objc public static let AccountTypeProUnlimited = "pro_unlimited"
+
```
ghking: Why are we adding these? Is this part of the conversion?

VimeoNetworking/Sources/NSError+Extensions.swift
```diff
@@ -83,7 +83,7 @@ public extension NSError
      
      - returns: a new `NSError`
      */
-    class func error(withDomain domain: String?, code: Int?, description: String?) -> NSError
+    @nonobjc class func error(withDomain domain: String?, code: Int?, description: String?) -> NSError
```
ghking: Aren't things `@nonobjc` by default? Do we need to add this ever?

VimeoNetworking/Sources/NSError+Extensions.swift
```diff
@@ -161,7 +161,7 @@ public extension NSError
         
         if let userInfo = userInfo
         {
-            augmentedInfo.append(userInfo)
+            augmentedInfo.append(userInfo as! Dictionary<String, Any>)
```
ghking: Rather than force casting here, we might want to reconsider the function argument to be `[String: Any]` rather than `[AnyHashable: Any]`

VimeoNetworking/Sources/Request.swift
```diff
@@ -169,7 +169,7 @@ public struct Request<ModelType: MappableResponse>
         {
             let queryString = AFQueryStringFromParameters(parameters)
             
-            if queryString.characters.count > 0
+            if queryString.count > 0
```
ghking: ⚔️

VimeoNetworking/Sources/ResponseCache.swift
ghking: @jasonhawkins Isn't that why you would use `@testable import` rather than `import` in the test files?

VimeoNetworking/Sources/Constants.swift
jasonhawkins: The compiler throws an error when we try to access this constant as a default argument 
in our **AppConfiguration** initializer. 

Here's the error:
```
AppConfiguration.swift:53:54: Let 'VimeoDefaultAPIVersionString' is internal and cannot be referenced from a default argument value
```

And here's the initializer:

```Swift
public init(clientIdentifier: String,
            clientSecret: String,
            scopes: [Scope],
            keychainService: String,
            keychainAccessGroup: String? = nil,
            apiVersion: String = VimeoDefaultAPIVersionString,
            baseUrl: URL = VimeoBaseURL)
{
    self.clientIdentifier = clientIdentifier
    self.clientSecret = clientSecret
    self.scopes = scopes
    self.keychainService = keychainService
    self.keychainAccessGroup = keychainAccessGroup
    self.apiVersion = apiVersion
    self.baseUrl = baseUrl
}
```

VimeoNetworking/Sources/ExceptionCatcher+Swift.swift
jasonhawkins: **ObjC_ExceptionCatcher** is Objective-C, of course. When we subclass in Swift we end up with the default access of`internal`, which makes sense. And as you pointed out below this should be elevatated to `open` when using `@testable` but… it's not. 

I've tried working around this but the best I could come up with was to just mark it as `public`. Maybe I'm missing something? Let me know if you want chat about it.

VimeoNetworking/Sources/Models/VIMLiveChatUser.swift
jasonhawkins: Ah I think I had brought these in around the time I was working on the live chat stuff. We shouldn't need them anymore.

VimeoNetworking/Sources/NSError+Extensions.swift
jasonhawkins: You would think, right? It looks like Foundation subclasses like this require adding this attribute. Here's the compiler error I was getting:

```
NSError+Extensions.swift:86:16: Method cannot be in an @objc extension 
of a class (without @nonobjc) because the type of the parameter 2 cannot
be represented in Objective-C
```

VimeoNetworking/Sources/NSError+Extensions.swift
jasonhawkins: I *think* this is a side effect of AFNetworking? Or at least the version that we're still using underneath everything. I would love to fix this but I think if you follow the chain up it ends up not being  a quick one.

VimeoNetworking/Sources/NSError+Extensions.swift
ghking: Ah gotcha

VimeoNetworking/Sources/NSError+Extensions.swift
ghking: @jasonhawkins I did a project-wide find and replace of `[AnyHashable: Any]` -> `[String: Any]`, and it worked out. 

I think this would probably be a good change to make since entities that we're using as `[AnyHashable: Any]` in this framework are either JSON or `NSError` user info. In both of these cases, the key has to be a String. It's the JSON spec, and even `NSError` types `userInfo` as `NSDictionary<NSErrorUserInfoKey, id>`, where `NSErrorUserInfoKey` is a typedef for `NSString`.

VimeoNetworking/Sources/ResponseCache.swift
ghking: https://www.natashatherobot.com/swift-2-xcode-7-unit-testing-access/

VimeoNetworking/Sources/VimeoClient.swift
```diff
@@ -174,7 +174,7 @@ final public class VimeoClient
      
      - returns: a `RequestToken` for the in-flight request
      */
-    public func request<ModelType: MappableResponse>(_ request: Request<ModelType>, completionQueue: DispatchQueue = DispatchQueue.main, completion: @escaping ResultCompletion<Response<ModelType>>.T) -> RequestToken
+    public func request<ModelType>(_ request: Request<ModelType>, completionQueue: DispatchQueue = DispatchQueue.main, completion: @escaping ResultCompletion<Response<ModelType>>.T) -> RequestToken
```
ghking: yesssss

VimeoNetworking/Sources/ResponseCache.swift
jasonhawkins: I know how this is _supposed_ to work, but if this class isn't public it won't be accessible from the test target. Even though it should be. The `@testable` attribute is supposed to allow us do exactly that, but it's not. Here's Apple's documentation:

> When you write an app with a unit test target, the code in your app needs to be made available to that module in order to be tested. By default, only entities marked as open or public are accessible to other modules. However, a unit test target can access any internal entity, if you mark the import declaration for a product module with the @testable attribute and compile that product module with testing enabled.

For our part we're using `@testable` for the import declaration and the product module does have testing enabled. 

I looked into this and created a new example class and example test case. I tried out the different access levels and ran into the exact same issue. Even though I was using `@testable`, internal classes were not accessible from the test case. BUT! I have a solution…

As of CocoaPods 1.3.0 we can move all of our tests _directly into the pod_. This seems like a pretty good option. I've got a WIP branch that demonstrates what that world might look like, built with CocoaPods 1.5.0: https://github.com/vimeo/VimeoNetworking/tree/tech/swift-4-cocoapods-1-5-0. And I've got a document with some of the high level changes between CocoaPods 1.3.1, 1.4.0, and 1.5.0 here: https://paper.dropbox.com/doc/CocoaPods-1.3.1-1.4.0-1.5.0-vDVpGmtp4IJKFoFtia9Ii.

What I'd like to do is either disable the tests that fail (temporarily) because of this access control issue, or change the access level of relevant classes (again, temporarily) for this PR and then work on getting us up-to-date with CocoaPods 1.5.0 and merge those changes too.

VimeoNetworking.podspec
```diff
@@ -6,13 +6,12 @@ Pod::Spec.new do |s|
   s.description  = "An iOS/tvOS library for interacting with the Vimeo API."
   s.homepage     = "https://github.com/vimeo/VimeoNetworking"
   s.license      = { :type => "MIT", :file => "LICENSE" }
-  s.authors      = { "Alfie Hanssen" => "alfiehanssen@gmail.com",
```
ghking: rip

VimeoNetworking.podspec
```diff
@@ -6,13 +6,12 @@ Pod::Spec.new do |s|
...
-                     "Rob Huebner" => "robh@vimeo.com",
```
ghking: rip

VimeoNetworking/Sources/Constants.swift
ghking: @jasonhawkins I'm thinking more like 

```swift 
 /// Base URL for the Vimeo API
let VimeoBaseURL = URL(string: "https://api.vimeo.com")!

 /// Default API version to use for requests
let VimeoDefaultAPIVersionString = "3.2"
```

so keeping those internal, and the init would look like:

```swift
public init(clientIdentifier: String,
                clientSecret: String,
                scopes: [Scope],
                keychainService: String,
                keychainAccessGroup: String? = nil,
                apiVersion: String? = nil,
                baseUrl: URL? = nil)
    {
        self.clientIdentifier = clientIdentifier
        self.clientSecret = clientSecret
        self.scopes = scopes
        self.keychainService = keychainService
        self.keychainAccessGroup = keychainAccessGroup
        self.apiVersion = apiVersion ?? VimeoDefaultAPIVersionString
        self.baseUrl = baseUrl ?? VimeoBaseURL
    }
```

VimeoNetworking/Sources/KeychainStore.swift
```diff
@@ -121,9 +121,9 @@ final class KeychainStore
     
     // MARK: - 
     
-    private func query(for key: String) -> [AnyHashable: Any]
+    private func query(for key: String) -> [String: Any]
     {
-        var query: [AnyHashable: Any] = [:]
+        var query: [String: Any] = [:]
```
ghking: dopeeeeee

jasonhawkins:
Ready for another look @ghking!

VimeoNetworking/Sources/AppConfiguration.swift
```diff
@@ -24,50 +24,41 @@
...
     {
         self.clientIdentifier = clientIdentifier
         self.clientSecret = clientSecret
         self.scopes = scopes
         self.keychainService = keychainService
         self.keychainAccessGroup = keychainAccessGroup
-        self.apiVersion = apiVersion
-        self.baseUrl = baseUrl
+        self.apiVersion = apiVersion ?? VimeoDefaultAPIVersionString
+        self.baseUrl = baseUrl ?? VimeoBaseURL
```
ghking: nice

PR #229, opened 2018-04-26
Changes VimeoBaseURL constant access back to public

jasonhawkins:
After some deliberation we decided to restore access to the VimeoBaseURL outside of the VimeoNetworking module 👍


jenniferlim:
🚀 @jasonhawkins  🚀

PR #270, opened 2018-10-10
VIM-XXX: Make sessionManager configurable

nicolelehrer:
#### Ticket

N/A

#### Issue Summary

- Provides a block to configure the session manager

#### Implementation Summary

- new closure type for session manager configuration
- pass in parameter of this type on api consuming session manager
- updates unit tests
- updates documentation

#### How to Test

Make sure unit tests pass and project builds


VimeoNetworking/Sources/VimeoClient.swift
```diff
@@ -469,10 +467,13 @@ extension VimeoClient
...
     ///   - appConfiguration: An AppConfiguration instance
-    public static func configureSharedClient(withAppConfiguration appConfiguration: AppConfiguration)
+    ///   - configureSessionManagerBlock: a block to configure the session manager
+    public static func configureSharedClient(withAppConfiguration appConfiguration: AppConfiguration, configureSessionManagerBlock: ConfigureSessionManagerBlock?)
     {
         self._sharedClient.configuration = appConfiguration
-        self._sharedClient.sessionManager = VimeoSessionManager.defaultSessionManager(appConfiguration: appConfiguration)
+        
+        let defaultSessionManager = VimeoSessionManager.defaultSessionManager(appConfiguration: appConfiguration, configureSessionManagerBlock: configureSessionManagerBlock)
+        self._sharedClient.sessionManager = defaultSessionManager
```
supperspidey: It looks like we're already done with the previous session manager. Do we want to invalidate it here, before using a new session manager?

nicolelehrer:
Good call out @supperspidey - I went with allowing old tasks to finish assuming there's a another actor in this who will handle invalidation later once all tasks finish, even if we reset the value of `self._sharedClient.sessionManager` immediately after calling this. In other words, I'm assuming that resetting the value of `self._sharedClient.sessionManager` right away won't undo our deed of trying to invalidate the session (if we're waiting on tasks).  But, if you think we just go ahead and cancel old tasks so we know session is invalidated right away, I'm ok with that approach too.

PR #294, opened 2019-02-08
Revert "VIM-6828: Add free trial copy to purchase flow UI"

Jumhyn:
Per feedback from Mike, we should avoid the version check at the network layer.


nicolelehrer:
@Jumhyn sounds good - will there be a client PR moving the check to call site? Or a wrapper method in client that gets used at all call sites?

Jumhyn:
@nicolelehrer Moving the check to a helper method in StoreController. Purchase flow PR addresses this.

PR #336, opened 2019-07-15
Adding placeholder to OAuth process

DavidBemerguy:
#### Pull Request Checklist

#### Issue Summary

- OAuth process on web implemented three new parameters that can be passed as an url query in order to facilitate user login/join
1. Email -> A placeholder for email in the email field
2. Name -> A placeholder for name in the name field
3. Auth action -> Declare if user should see first the login screen or the join one

#### Implementation Summary

- We added three parameters in the method codeGrantAuthorizationURL in the AuthenticationController class. The method is now returning an url which include those parameters 

#### Reviewer Tips

Mainly the codeGrantAuthorizationURL method in the AuthenticationController class was modified.
Another helper method was added

#### How to Test

1. Unit test was added and should pass
2. Add parameters to codeGrantAuthorizationURL method and check if, when Safari is opened, the placeholders are in place

### DIFF
--- a/Sources/Shared/AuthenticationController.swift
+++ b/Sources/Shared/AuthenticationController.swift
@@ -584,8 +584,11 @@ private extension URL {
         let newQueries = urlQueries.compactMap({ (item) -> URLQueryItem? in
             return item.value != nil ? item : nil
         })
-        queries.append(contentsOf: newQueries)
-        urlComponents?.queryItems = queries
+        if newQueries.count > 0 {
+            queries.append(contentsOf: newQueries)
+            urlComponents?.queryItems = queries
+        }
+    
         guard let newUrl = urlComponents?.url else {
             fatalError("Could not generate queries on code grant auth URL. Probably some invalid characters were added")
         }
--- a/Sources/Shared/AuthenticationController.swift
+++ b/Sources/Shared/AuthenticationController.swift
@@ -587,7 +587,7 @@ private extension URL {
         queries.append(contentsOf: newQueries)
         urlComponents?.queryItems = queries
         guard let newUrl = urlComponents?.url else {
-            return self
+            fatalError("Could not generate queries on code grant auth URL. Probably some invalid characters were added")
         }
         return newUrl
     }
--- a/Sources/Shared/AuthenticationController.swift
+++ b/Sources/Shared/AuthenticationController.swift
@@ -184,7 +184,7 @@ final public class AuthenticationController {
     ///   - email: Optional placeholder: when set, the user will see a placeholder email inside the email text field in Safari
     ///   - action: Set this parameter to send user straight to join or login inside Safari, default is login
     /// - Returns: the code grant authorization page URL
-    public func codeGrantAuthorizationURL(usingPlaceholdersForName name: String? = nil, andEmail email: String? = nil, showingInitialAuthorizationAction action: AuthorizationAction = .login) -> URL {
+    public func codeGrantAuthorizationURL(usingPlaceholdersForName name: String? = nil, andEmail email: String? = nil, showingInitialAuthorizationAction action: AuthorizationAction? = nil) -> URL {
         let parameters = [Constants.ResponseTypeKey: Constants.CodeKey,
                           Constants.ClientIDKey: self.configuration.clientIdentifier,
                           Constants.RedirectURIKey: self.codeGrantRedirectURI,
@@ -195,7 +195,7 @@ final public class AuthenticationController {
             .appendingQueries([
                 URLQueryItem(name: Constants.URLQueries.Email, value: email),
                 URLQueryItem(name: Constants.URLQueries.Name, value: name),
-                URLQueryItem(name: Constants.URLQueries.AuthorizationAction, value: action.rawValue)
+                URLQueryItem(name: Constants.URLQueries.AuthorizationAction, value: action?.rawValue)
                 ])
             .absoluteString
         
--- a/Sources/Shared/AuthenticationController.swift
+++ b/Sources/Shared/AuthenticationController.swift
@@ -47,6 +47,17 @@ final public class AuthenticationController {
         static let ScopeKey = "scope"
         static let StateKey = "state"
         static let CodeGrantAuthorizationPath = "oauth/authorize"
+        
+        fileprivate struct URLQueries {
+            static let Email = "email"
+            static let Name = "name"
+            static let AuthorizationAction = "auth_action"
+        }
+    }
+    
+    public enum AuthorizationAction : String {
+        case join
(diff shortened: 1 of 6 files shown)


Sources/Shared/AuthenticationController.swift
```diff
@@ -166,29 +177,38 @@ final public class AuthenticationController {
...
                           Constants.RedirectURIKey: self.codeGrantRedirectURI,
                           Constants.ScopeKey: Scope.combine(self.configuration.scopes),
                           Constants.StateKey: type(of: self).state]
         
-        let urlString = self.configuration.baseUrl.appendingPathComponent(Constants.CodeGrantAuthorizationPath).absoluteString
+        let urlString = self.configuration.baseUrl.appendingPathComponent(Constants.CodeGrantAuthorizationPath)
+            .appendingQueries([
+                URLQueryItem(name: Constants.URLQueries.Email, value: email),
+                URLQueryItem(name: Constants.URLQueries.Name, value: name),
+                URLQueryItem(name: Constants.URLQueries.AuthorizationAction, value: action.rawValue)
```
rpassis: Because we always have a default value for `login` this is changing the current behavior in that the URL returned from `codeGrantAuthorization()` will now always include the `auth_action` query item. 

How can we ensure that this change will not affect the existing authorization flow?

Sources/Shared/AuthenticationController.swift
```diff
@@ -166,29 +177,38 @@ final public class AuthenticationController {
...
         
         var error: NSError?
         let urlRequest = VimeoRequestSerializer(appConfiguration: self.configuration).request(withMethod: VimeoClient.Method.GET.rawValue, urlString: urlString, parameters: parameters, error: &error)
         
-        guard let url = urlRequest.url, error == nil
-        else {
-            fatalError("Could not make code grant auth URL")
+        guard let codeGrantAuthorizationURL = urlRequest.url, error == nil
+            else {
+                fatalError("Could not make code grant auth URL")
```
rpassis: It's probably worth adding tests for invalid URL characters to see how this code path would behave in your unit tests. As far as I can see, if the call to `appendingQueries` failed somehow, the URL returned is the unmodified version of itself, which would also exclude `auth_action`. Is this acceptable?

Tests/Shared/AuthenticationControllerTests.swift
```diff
@@ -0,0 +1,58 @@
...
+            XCTFail("Unable to retrieve components from code grant url")
+            return
+        }
+        
+        guard let queryItems = urlComponents.queryItems else {
+            XCTFail("No query items available in code grant url")
+            return
+        }
+        
+        let authAction = queryItems.first { $0.name == "auth_action" }
```
rpassis: Can you please add tests for name and email, as well as invalid URL characters?

rpassis (changes requested):
Hey @DavidBemerguy thanks for the PR, it's looking pretty good!

I left a few comments for you to consider - let me know what you think and if you have any questions.

Rog

rpassis:
@DavidBemerguy By the way I have merged develop into your branch so you will need to pull from remote first. There's also a project merge conflict that will need to be fixed at some point. Feel free to rebase and push it back up if you prefer!

Sources/Shared/AuthenticationController.swift
DavidBemerguy: Agree, we need to let the user choose what to add.

Sources/Shared/AuthenticationController.swift
DavidBemerguy: We should think about fallback, we would return unmodified URL and the user won't know why the placeholder is not shown. Any idea? I though on this solution because it keeps the same API. I can't get right now to a better solution

Tests/Shared/AuthenticationControllerTests.swift
DavidBemerguy: Fore sure

Sources/Shared/AuthenticationController.swift
DavidBemerguy: So I though in case the user passed a name or email with invalid characters we should also make a fatalerror since I want to keep the same API and let the user know that some of the queries are invalid

Sources/Shared/AuthenticationController.swift
```diff
@@ -555,3 +575,24 @@ final public class AuthenticationController {
         self.client.currentAccount = account
     }
 }
+
+// MARK: Helpers
+private extension URL {
+    func appendingQueries(_ urlQueries: [URLQueryItem]) -> URL {
+        var urlComponents = URLComponents(url: self, resolvingAgainstBaseURL: true)
+        var queries: [URLQueryItem] = urlComponents?.queryItems ?? []
```
rpassis, line 583: can you please rename this do `existingQueries` to make it clearer that we are dealing with queries that may already be part of the url?

Sources/Shared/AuthenticationController.swift
```diff
@@ -555,3 +575,24 @@ final public class AuthenticationController {
...
+        let newQueries = urlQueries.compactMap({ (item) -> URLQueryItem? in
```
rpassis, line 584: `filter` is probably more appropriate here and would read better:

`queries.filter { $0.value != nil }`

And you also simplify the logic here by just adding existing and new queries, no need for count check etc:

`urlComponents.queryItems = existingQueries + newQueries`

Sources/Shared/AuthenticationController.swift
```diff
@@ -555,3 +575,24 @@ final public class AuthenticationController {
         self.client.currentAccount = account
     }
 }
+
+// MARK: Helpers
+private extension URL {
+    func appendingQueries(_ urlQueries: [URLQueryItem]) -> URL {
+        var urlComponents = URLComponents(url: self, resolvingAgainstBaseURL: true)
```
rpassis, line 582: Perhaps we could try to unwrap `urlComponents` here and return `self` if we are unable to?

`guard let comps = URLComponents(...) else { return self }`

This way you avoid all the unwrapping that follows, as well as the possibility of fatalError'ing.

As far as I can see there is no harm done in returning the original URL instead of a runtime crash as that would still allow the user to continue with the authentication process, just without some of the convenience placeholders.

rpassis (changes requested):
Hi @DavidBemerguy I've left some minor comments for you to take a look.
Note that there are some scheme changes that should not be part of this commit so we will have to get these reverted before merging.
Let me know if you have any questions!
Thanks,
Rog

rpassis:
Closing this due to inactivity @DavidBemerguy. Feel free to open a new PR if you wish to continue working on it.

PR #359, opened 2019-09-30
VIM-7293: Changes default JSON serializer behavior on VimeoSessi…

rpassis:
#### Ticket

[VIM-7293](https://vimean.atlassian.net/browse/VIM-7293)

#### Pull Request Checklist

- [X] Resolved any merge conflicts
- [X] No build errors or warnings are introduced
- [X] New files are written in Swift
- [X] New classes contain license headers
- [X] New classes have Documentation
- [X] New public methods have Documentation

#### Issue Summary

Currently the `VimeoSessionManager` that inherits from `AFHTTPURLSessionManager` and handles requests on VimeoClient's behalf is tightly coupled with both the request and response serializers provided by the library.

Basically every time a request is made the data returned with the response is automatically serialized by the internal of `AFNetworking` and returned to us as a JSON object, and there is no way to opt-out of this behavior.

The goal of this ticket is to make the required changes so that we can opt-in to JSON serialization of responses as appropriate, offering the ability to receive raw data when needed (effectively bypassing the JSON response serializer), while maintaining backwards compatibility with the existing API and its consumers.

This will also allow us to support `Decodable` types and eventually remove our dependency on the dynamic object mapping.

#### Implementation Summary

The important changes in this PR are inside `VimeoSessionManager`.

Firstly, `VimeoSessionManager` no longer inherits from `AFHTTPSessionManager`. This change is mostly cosmetic at the moment as internally we still use an instance of `AFHTTPSessionManager` to manage requests.

Previously the session manager was initialized with a default JSON request & response serializers (`VimeoRequestSerializer` and `VimeoResponseSerializer`). This was changed so the default AFNetworking serializers are used instead.

What this effectively means is that both requests and responses using `VimeoSessionManager` are now "opt-in" serializable. We can also opt to receive raw `Data` types from the response, allowing us to opt-in to the appropriate serialization depending on the calling site and public API used.

The next change once the above was in place was to add the public APIs to allow the calling site to choose how they want the response to be treated. The current options are:

- Raw `Data`
- JSON serialization using `VimeoResponseSerializer` (previous default behavior)
- Decodable types using `JSONDecoder`
- Uploads & downloads

In addition to the above I have also refactored the public API for better ergonomics and stricter, more meaningful responses using `Result` types.

There were also some cosmetic changes and addition of wrapper methods on `VimeoSessionManager`. The main reason for this is that since we no longer inherit from `AFHTTPSessionManager` I decided to make the underlying session manager internal to the framework and only expose whatever methods the clients (Upload and iOS) already use via the session manager.

#### How to Test

- Build and run, ensure all unit tests pass
- On the client side, build and run [this PR](https://github.vimeows.com/MobileApps/Vimeo-iOS/pull/3091) and ensure the app behaves as expected, including downloading and uploading videos.


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">SwiftLint not installed, download from https://github.com/realm/SwiftLint</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 173 tests, with 0 failures (0 unexpected) in 1.129 (1.291) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1373)

</p>

rpassis:
Hey @nicolelehrer @Jumhyn this is ready for a good look when you get a chance. I assigned Nicole as the primary PR reviewing but think a second pair of eyes on this would be very useful if you have time next week Freddy!

Sources/Shared/Core/VNError.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  VimeoNetworkingError.swift
+//  VNError.swift
```
bryantjustin: Not a huge fan of the name, but I also can't think of another name that isn't as cumbersome as the existing one.

Sources/Shared/VimeoClient.swift
```diff
@@ -27,67 +27,62 @@
...
-         */
+        public let task: Task?
+
+        /// Cancels the request
         public func cancel() {
             self.task?.cancel()
         }
     }
     
-        /// Dictionary containing URL parameters for a request
```
bryantjustin, line 54: 🙏. These indentation issues have been bugging me for awhile.

Tests/Shared/VIMUserTests.swift
```diff
@@ -37,7 +37,7 @@ class VIMUserTests: XCTestCase {
     override func setUp() {
         super.setUp()
         
-        VimeoClient.configureSharedClient(withAppConfiguration: AppConfiguration(clientIdentifier: "{CLIENT_ID}",
+        VimeoClient.configure(with: AppConfiguration(clientIdentifier: "{CLIENT_ID}",
```
bryantjustin: The indentation looks like it's gonna be messed up in this file. I'm assuming you ran the rename refactor? Are we gonna lint/format this project at some point?

Sources/Shared/VimeoClient.swift
```diff
@@ -27,67 +27,62 @@
...
+
+        static let pagingKey = "paging"
+        static let totalKey = "total"
+        static let pageKey = "page"
+        static let perPageKey = "per_page"
+        static let nextKey = "next"
+        static let previousKey = "previous"
+        static let firstKey = "first"
+        static let lastKey = "last"
     }
```
bryantjustin: Worth moving these to an extension at the bottom of the file? 

Also should we maybe split them into separate types:
```swift
enum ResponseKeys {
    static let paging = "paging"
    static let total = "total"
    ...
}

enum AuthHeaderConstants { }
```

Or even simply add them as constants to a `String` extension?

Sources/Shared/VimeoClient.swift
```diff
@@ -145,136 +140,214 @@ final public class VimeoClient {
...
+            let _ = self._sharedInternal.sessionManager else {
+                assertionFailure("VimeoClient.sharedClient must be configured before accessing")
+                return self._sharedInternal
         }
-        else {
-            let task = self.sessionManager?.request(with: request, then: { response in
+        return self._sharedInternal
+    }
+
+    private static let _sharedInternal = VimeoClient()
```
bryantjustin: I think it would be fine to just call this `_shared` no? I thought maybe that was a common convention.

Sources/Shared/Core/VNError.swift
```diff
@@ -8,10 +8,14 @@
 
 import Foundation
 
-enum VimeoNetworkingError: Error {
+public enum VNError: Error {
     case encodingFailed(EncodingFailedReason)
+    case invalidURL
+    case requestError(Error)
+    case serializatingError
```
bryantjustin: Did you mean:
```suggestion
    case serializationError
```
or edit
```suggestion
    case serializingError
```
?

bryantjustin (changes requested):
Mostly looks good, I still need to run and test the iOS repo, but I do have a few questions in the mean time.

The only one that I would really like to see resolved is the naming of `VNError.serializatingError`. Everything else is kinda low priority/severity IMO. I've proposed alternatives, but not sure which one you actually meant.

Sources/Shared/Core/VNError.swift
rpassis: Yeah I agree I don't love it either, will see if I can come up with something else 🤔

Sources/Shared/VimeoClient.swift
rpassis: One of the response keys is currently used outside of this file so that stopped me from adding them as private extensions on String. But I agree separating them and moving down the bottom of the file makes sense!

rpassis: Actually nevermind I think I can sort this out.

Sources/Shared/VimeoClient.swift
rpassis: Yes I should've reverted this change, this was of something I decided to do on a separate PR which is remove the singleton from the library completely and let the consumer decide whether it wants to set it up as a singleton. There's no reason the networking library should be making that decision.

Sources/Shared/Core/VNError.swift
rpassis: @bryantjustin I'll leave the naming as is unless you feel strongly about it. If I can think of a better option before this is merged into develop I'll change it on a separate PR.

bryantjustin (approved):
Great job! Upload and delete worked fine. This brings us one step closer to an AFNetworking-less life.

PR #364, opened 2019-10-09
VIM-6900: Introduces changes to support deeper interaction with Showcases

jasonhawkins:
### Ticket
[VIM-6900](https://vimean.atlassian.net/browse/VIM-6900)

### Implementation Summary
Introduces support for creating, editing and deleting a showcase. Also adds support for batch-adding videos to a showcase, and adding a video to multiple showcases.

### How to Test
This work has been reviewed previously but feedback is still welcome.


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">SwiftLint not installed, download from https://github.com/realm/SwiftLint</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 118 tests, with 0 failures (0 unexpected) in 1.201 (1.380) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1361)

</p>

Sources/Shared/BatchOperable.swift
```diff
@@ -0,0 +1,53 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+/// Conforming objcets are capable of being added to, or removed from, a containing object in a single request.
```
rpassis: `objects`

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+    static let privacy = "privacy"
+}
+
 public extension Request {
-    
     /// Returns a new request to fetch a specific album.
     ///
     /// - Parameter uri: The album's URI.
     /// - Returns: Returns a new `Request` for an individual album.
     static func albumRequest(for uri: String) -> Request {
```
rpassis, line 38: The naming here seems ambiguous, is this a `fetchAlbum` request? 
Also do we always append request to the method names in a Request type? Seems redundant from the calling site i.e. `Request.createAlbum(...)` instead of `Request.createAlbumRequest(...)`

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
         return Request(path: uri)
     }
+
+    /// Returns a new request for creating an album.
+    /// - Parameter userURI: The URI for the current user, and owner of the album being created.
+    /// - Parameter name: The name of the album being created.
+    /// - Parameter description: An optional description for the album.
+    /// - Parameter privacy: The privacy parameter as a String. If none is specified it defaults to "anyone".
+    /// - Parameter password: An optional password parameter, only required when privacy is set to "password".
+    static func createAlbumRequest(
```
rpassis, line 48: I'm not sure how the `Album` model is setup but would it make sense to pass a locally constructed `Album` here instead of its individual attributes? Depending on use case this list of parameters could get pretty long!

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+        name: String,
+        description: String? = nil,
+        privacy: String? = VIMPrivacy_Public,
+        password: String? = nil
+    ) -> Request {
+        var parameters = [String: String]()
+        parameters[Key.name] = name
+        parameters[Key.description] = description
+
+        privacy.map { parameters[Key.privacy] = $0 }
```
rpassis: why the mapping here? Can't you just use the subscript directly how you did with `description`?

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+        return Request(method: .POST, path: userURI + "/albums", parameters: parameters)
+    }
+
+    /// Returns a new request for updating an exising album.
+    /// - Parameter albumURI: The URI for the album that will be updated.
+    /// - Parameter name: The name of the album.
+    /// - Parameter description: An optional description for the album.
+    /// - Parameter privacy: The privacy parameter as a String.
+    /// - Parameter password: An optional password parameter, only required when the privacy is set to "password".
+    static func updateAlbumRequest(
```
rpassis, line 70: Ditto re: naming

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+        name: String,
+        description: String? = nil,
+        privacy: String?,
+        password: String? = nil
+    ) -> Request {
+        var parameters = [String: String]()
+        parameters[Key.name] = name
+        parameters[Key.description] = description
+
+        privacy.map { parameters[Key.privacy] = $0 }
```
rpassis, line 80: same comments as above apply, these 2 function could pretty much be folded into a private one that gets called with a different `method` and `path` depending on the intent.

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+    /// - Parameter userURI: The URI for the current user, and owner of the album being created.
+    /// - Parameter name: The name of the album being created.
+    /// - Parameter description: An optional description for the album.
+    /// - Parameter privacy: The privacy parameter as a String. If none is specified it defaults to "anyone".
+    /// - Parameter password: An optional password parameter, only required when privacy is set to "password".
+    static func createAlbumRequest(
+        userURI: String,
+        name: String,
+        description: String? = nil,
+        privacy: String? = VIMPrivacy_Public,
```
rpassis: is there a need for this to be optional if you are providing a default? 
Should I ever be allowed to set the privacy of my Album to nil?

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+    /// - Parameter albumURI: The URI for the album that will be updated.
+    /// - Parameter name: The name of the album.
+    /// - Parameter description: An optional description for the album.
+    /// - Parameter privacy: The privacy parameter as a String.
+    /// - Parameter password: An optional password parameter, only required when the privacy is set to "password".
+    static func updateAlbumRequest(
+        albumURI: String,
+        name: String,
+        description: String? = nil,
+        privacy: String?,
```
rpassis, line 74: Same issue with privacy here though it is trickier in this case as you may not want to update that specific field.

Sources/Shared/Request+Album.swift
```diff
@@ -29,13 +29,71 @@ public typealias AlbumRequest = Request<Album>
...
+
+        privacy.map { parameters[Key.privacy] = $0 }
+        password.map { parameters[Key.password] = $0 }
+
+        return Request(method: .PATCH, path: albumURI, parameters: parameters)
+    }
+
+    /// Returns a new request to delete the album for the given URI.
+    /// - Parameter albumURI: The album's URI.
+    static func deleteAlbumRequest(for albumURI: String) -> Request {
```
rpassis, line 88: this and all of the methods above would definitely read a lot better if we could just pass the Album object directly

```swift
let request = Request.delete(album)
```

Sources/Shared/Models/VIMInteraction.m
```diff
@@ -93,4 +89,20 @@ - (void)setStreamStatus
     self.streamStatus = [number intValue];
 }
 
+- (BOOL)canGet {
+    return (self.options && [self.options containsObject:@"GET"]);
```
rpassis, line 93: can we use the `HTTPMethod` string values here instead of hard coding our own?

Sources/Shared/Models/Album.swift
```diff
@@ -86,8 +87,9 @@ import Foundation
     @objc public var videoThumbnails: NSArray?
     @objc public var user: VIMUser?
     @objc public var theme: String?
-    @objc dynamic private var metadata: [String: Any]?
-    @objc dynamic private var connections: [String: Any]?
+    @objc private var metadata: [String: Any]?
```
rpassis, line 90: is there a reason why we had these set as `dynamic` before?

Sources/Shared/Models/Album.swift
```diff
@@ -130,11 +132,16 @@ import Foundation
         self.createdTime = self.formatDate(from: self.createdTimeString)
         self.modifiedTime = self.formatDate(from: self.modifiedTimeString)
         self.parseConnections()
+        self.parseInteractions()
     }
     
-    public func connectionWithName(connectionName: String) -> VIMConnection? {
+    public func connection(withName connectionName: String) -> VIMConnection? {
```
rpassis: consider using `connection(named:)` instead?

Sources/Shared/Models/Album.swift
```diff
@@ -130,11 +132,16 @@ import Foundation
...
         return self.connections?[connectionName] as? VIMConnection
     }
+
+    public func interaction(withName interactionName: String) -> VIMInteraction? {
```
rpassis: as above `interaction(named:)`

Sources/Shared/Models/Album.swift
```diff
@@ -156,6 +161,17 @@ import Foundation
             }
         }
     }
+
+    private func parseInteractions() {
+        guard let dictionary = self.metadata?[Constant.Key.Interactions] as? [String: Any] else { return }
+
+        self.interactions = [String: Any]()
```
rpassis: Can I suggest that you create a local dictionary here, set it up with the interactions you find and then assign it to the property after the for-loop? 

The reason I mention is that I wonder how safe it is to modify this property in this for loop directly in the event that this somehow is called concurrently from multiple threads? It would still not be 100% safe but at tiny bit safer, what do you think?

Sources/Shared/BatchOperable.swift
rpassis: Is it straightforward to add an example of how adopting types could take advantage of this? It's hard to get what this does without a concrete use case.

rpassis (changes requested):
Hey @jasonhawkins left some comments for you, mostly cosmetic but hopefully it all makes sense.

Sources/Shared/Request+Album.swift
jasonhawkins: I agree that the naming is weird and could be improved, but it's consistent with how (most) of our other request methods are named right now.

Sources/Shared/Request+Album.swift
jasonhawkins: This is an implementation detail in the app, but the presenter we're working with is only aware of a view-model, and we're relying on the server to return an actual `Album` object if the `.POST` is successful.

Sources/Shared/Request+Album.swift
jasonhawkins: It's probably not strictly necessary to make the privacy option optional, but you _can_ actually ask the server to create an album without setting its privacy in the request, and the server default will also be to set the album to "anybody". 

So here you can omit the parameter and we'll set the privacy to "anybody". Or you can set it to `nil` and the server will set it to "anybody." Both are valid.

In the end I'm mapping here because sending the privacy parameter isn't a requirement.

Sources/Shared/Request+Album.swift
jasonhawkins: See above ^

Sources/Shared/Request+Album.swift
jasonhawkins: I did originally have just one request, but it felt like too much to ask the client to know what the correct endpoints were, to set the HTTP method themselves, and to come up with a good enough name that would apply to both creating and editing. 

There's also different behavior here in that omitting the `privacy` parameter won't ever send a default value.

Sources/Shared/Request+Album.swift
jasonhawkins: Ah yeah, see above ^

Sources/Shared/Request+Album.swift
jasonhawkins: See above ^

Sources/Shared/Request+Album.swift
jasonhawkins: I agree for the most part, and maybe we could discuss refactoring these after this phase is out? These are the problems though:
- It breaks with the convention we currently have for requests (even if it's weird and not necessarily a good one).
- Not all requestors (like the one for creation) have access to an `Album` object, though most would.
- If we're moving toward a world where our models are extracted from the networking layer we wouldn't want these requests to depend on a precise implementation of the album/showcase object. Maybe something like a protocol instead, but it feels like we're still a ways off from that.

Sources/Shared/Models/VIMInteraction.m
jasonhawkins: Yes and I think you just made a great case for rewriting this in Swift 😂. Right now our HTTPMethod constants are defined in a Swift enum.

Sources/Shared/Models/Album.swift
jasonhawkins: Not a good reason, no. I had a discussion with Alex about this a few months ago and it was likely a mistake. I couldn't think of a good reason why only these two properties would need to be dynamically dispatched when everything else isn't, yet all of them are accessible via Objective-C. 

A lot of our models are marked with `dynamic` which feels like maybe a mistake? Or do we really want to make sure we're using dynamic instead of static dispatch?

Sources/Shared/Models/Album.swift
jasonhawkins: Yes, good call. I'm going to do the same for connections above 👍

jasonhawkins:
Thanks for the feedback @rpassis, this is ready for another look.

Sources/Shared/Models/Album.swift
rpassis: yeah I can't really think of a reason this would be necessary.

Sources/Shared/Request+Album.swift
rpassis: So are you saying the parameter gets sent with a blank value if we set this to `nil`? I thought it would simply remove the key from the dictionary no?

jasonhawkins: You're correct, the key is removed from the dictionary entirely and no value is sent. The server accepts this and applies the default "anyone" privacy.

Jumhyn: Rog makes a good point! I think I suggested the `map` since the original logic was `if let privacy = privacy { parameters[Key.privacy] = privacy }`, but I didn't think about the fact that this logic wasn't necessary at all...

jasonhawkins: Should I just make the parameter non-optional and apply our default here, instead of relying on the server to apply the default? That would probably make things simpler here.

rpassis: I think I'd agree with that.

jasonhawkins: Okay cool, that sounds good to me.

rpassis (approved):
LGTM!

jasonhawkins:
build this please

PR #390, opened 2019-12-02
VIM-7487: Adds support for connected apps

jasonhawkins:
### Ticket
[VIM-7487](https://vimean.atlassian.net/browse/VIM-7487)

### Pull Request Checklist

### Ticket Summary
Update VimeoNetworking to include support for the new Connected Apps endpoints.

#### ConnectedApp
An object that contains all of the data related to a connection to a specific platform. However, not all of the properties relate to all platforms. Exceptions are noted in the documentation where applicable. Generally, `ConnectedApp` includes data related to when the connection was established, the platform it represents, and the scopes needed to interact with a particular feature. For now features are limited to live streaming and publishing to social.

#### Request+ConnectedApp
Adds convenience methods for constructing requests related to connected apps. This includes requesting all existing connected apps, creating a connection to a specific platform, and deleting an existing connection.

#### VIMConnection
Adds `VIMConnectionNameConnectedApps`, allowing a client to access the connected apps connection.

#### VIMInteraction
Expanded to support VIMUser interactions related to connected apps. You can now look for interactions on a user for each of the supported platforms (`facebook_connected_app`, for instance) to get information about whether the app is already connected, and what scopes are required for each of the supported features.

### Reviewer Tips
Testing this work requires access to the internal branch, `https://connected-apps-endpoints-api.ci.vimeows.com`. Furthermore, these features will only available to whitelisted apps for the foreseeable future once the changes are merged into the API.

### How to Test
Ensure unit tests pass. If testing against the CI branch ensure your client application is able to make the included requests and parse the results.


Sources/Shared/Request+ConnectedApp.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// - Parameters:
+    ///   - appType: The app platform for which the connection will be established.
+    ///   - token: An authentication token from the provided platfrom, used to establish the connection.
+    static func connect(to appType: ConnectedAppType, with token: String) -> Request {
+        let uri: String = .connectedAppsURI + appType.stringValue
+
+        var tokenKey: String
+        switch appType {
+        case .facebook:
+            tokenKey = .accessToken
```
antonklevets: https://github.vimeows.com/Vimeo-API/design/pull/146/files#diff-a9ed519c3fbb374ef99e26b80498642fR25
For facebook should use `.authCode`

antonklevets: PR Fix https://github.com/vimeo/VimeoNetworking/pull/392

cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 211 tests, with 0 failures (0 unexpected) in 1.011 (1.334) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1619)

</p>

Sources/Shared/Request+ConnectedApp.swift
jasonhawkins: Thanks, @antonklevets. I've updated this branch so that Facebook uses `auth_code` and updated the related test.

jasonhawkins:
build this please

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+/// - linkedin: Represents a connection to LinkedIn.
+/// - twitter: Represents a connection to Twitter.
+/// - youtube: Represents a connection to YouTube.
+@objc public enum ConnectedAppType: Int {
+    case facebook
+    case linkedin
+    case twitter
+    case youtube
+
+    public var stringValue: String {
```
rpassis: how about conforming this to `CustomStringConvertible` and changing the naming here to `description` so it conforms? I'm guessing it would need to be done in an extension to avoid the @objc annotation but that should be fine since this property is not exposed to Objective-C runtime.

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+
+///  An object that encapsulates the scopes necessary for interacting with features like publishing to social platforms
+///  or simulcasting a live stream.
+@objcMembers
+public class ConnectedAppScopes: VIMModelObject {
+
+    /// All scopes required for publishing to a specific social media platform.
+    public var publishToSocial: [String]?
```
rpassis: Is there a finite list of scopes that we can enumerate to make this a bit more type safe? At the moment I'm not even sure how to format this `var` with the different scopes - is it comma separate, spaces etc?

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+
+    /// All scopes required for simulcasting to a specific social media platform.
+    public var simulcast: [String]?
```
rpassis: same as above

Sources/Shared/Models/ConnectedApp.swift
rpassis: Also does it make sense for all of these properties to be publicly writeable as well? I'm not sure how they are going to be used but the less mutability of these models the better.

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+        ]
+    }
+}
+
+/// A category that can be sent when publishing to a social media platform.
+@objcMembers
+public class PublishOptionItem: VIMModelObject {
+
+    /// The ID of the publish item.
+    public var identifier: String?
```
rpassis: Same re: public setters - I'm guessing the identifier should not be publicly writeable?

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+
+    /// The name or display name of the publich item, i.e.: "art", "family", "vacation" etc.
```
rpassis: typo on `publish`

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+/// stream to multiple destinations, or publishing across platforms, require requesting specific scopes. The scopes
+/// required will always be returned in the `neededScopes` array.
+/// - Note: Some properties are specific to a particular platform. These cases have been noted in the documentation
+///         where relevant.
+@objc public class ConnectedApp: VIMModelObject {
+
+    /// The date when the user established this connected app, as a String.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `addDate` is preferred.
+    @objc public var addDateString: String?
```
rpassis: Same comment re: public setters

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,237 @@
...
+
+    /// The date when the user established this connected app.
+    @objc public var addDate: Date?
+
+    /// Facebook only. A value of 1 maps to expired data access. A value of 0 maps to having data access.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `isDataAccessExpired` is preferred.
+    @objc public var dataAccessIsExpired: NSNumber?
```
rpassis: does this need to be public given the computed boolean below?

Sources/Shared/Models/VIMInteraction.h
```diff
@@ -53,21 +59,33 @@ typedef NS_ENUM(NSInteger, VIMInteractionStreamStatus) {
...
-@property (nonatomic, copy, nullable) NSString *download;
-@property (nonatomic, strong, nullable) NSDate *expirationDate;
-@property (nonatomic, strong, nullable) NSDate *purchaseDate;
-@property (nonatomic, assign) VIMInteractionStreamStatus streamStatus;
+#pragma mark - Connected Apps
+/**
+ For connected app interactions this indicates whether the app is connected. A value of 0 or `nil` means no
+ connection has been established, or an existing one has expired.
+ */
+@property (nonatomic, strong, nullable) NSNumber *isConnected;
```
rpassis: could this be made private and exposed as a real (computed) boolean?

Sources/Shared/Requests/Request+ConnectedApp.swift
```diff
@@ -0,0 +1,76 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+public extension Request {
```
rpassis: should this be constrained to the `ConnectApp` type with `where ModelType ==`? I'm not sure how the type inference would work here without it - I'm guessing the consumer of the API would have to explicitly specify the request type otherwise?

Sources/Shared/Requests/Request+ConnectedApp.swift
```diff
@@ -0,0 +1,76 @@
...
+    /// - Parameter appType: The `ConnectedAppType` to fetch data for.
+    static func connectedApp(_ appType: ConnectedAppType) -> Request {
+        return Request(path: .connectedAppsURI + appType.stringValue)
+    }
+
+    /// Returns a `put` request for connecting the provided app type to the current authenticated user's account.
+    /// - Parameters:
+    ///   - appType: The app platform for which the connection will be established.
+    ///   - token: An authentication token from the provided platfrom, used to establish the connection.
+    static func connect(to appType: ConnectedAppType, with token: String) -> Request {
```
rpassis, line 44: Is it worth making this request more generic? As in do we always expect a token to be the method of authenticating with a third party? Could there be cases where we need to pass more than 1 parameter?

rpassis (changes requested):
This looks good Jason, I love how thoroughly tested it is! 

I've left a few comments, mostly regarding visibility & availability of certain properties for you to consider. Let me know what you think!

Sources/Shared/Models/ConnectedApp.swift
jasonhawkins: The actual scopes aren't defined in the client app, instead they come from the server. We get them by looking for an interaction on the user object for a platform, `facebook_connected_app`, for example. The returned `VIMInteraction` will contain the scopes needed for the supported "connected app" features.

This class is supposed to encapsulate and store those scopes on a `VIMInteraction` object. 

```ObjC
@interface VIMInteraction : VIMModelObject
…
@property (nonatomic, strong, nullable) ConnectedAppScopes *allScopes;`
```

But I think you're right about the access for these properties though. I'll take a look.

Sources/Shared/Models/ConnectedApp.swift
jasonhawkins: Good call, it does not.

Sources/Shared/Requests/Request+ConnectedApp.swift
jasonhawkins: For the platforms we support we only require the token for authentication. I guess it's possible that some future platform might have different requirements? But I think for now this should be fine.

jasonhawkins:
Thanks for the feedback @rpassis, this is ready for another look 👀

Sources/Shared/Requests/Request+ConnectedApp.swift
```diff
@@ -23,7 +23,7 @@
 //  THE SOFTWARE.
 //
 
-public extension Request {
+public extension Request where ModelType: ConnectedApp {
```
antonklevets: @jasonhawkins In this case model for `connectedApps()` method is wrong and always empty. There should be array of `ConnectedApp`

I used this code
```
let request = Request<[ConnectedApp]>.connectedApps()
....
case .success(let response):
   response.model // Array of ConnectedApp
```
and was all fine.

antonklevets: Also for `deleteConnectedApp` seems should be `VIMNullResponse` model instead of `ConnectedApp`, because there no info in response, i.e. i always receive error with message `VimeoClient requestSuccess returned invalid/absent dictionary`

should be 
`Request<VIMNullResponse>.deleteConnectedApp(.facebook)`

jasonhawkins: Hey Anton, you're totally right. Sorry about that. I've just pushed changes that will constrain the request objects to the correct types, and added some tests to validate those types.

You should now be able to use those request objects without having to specify `<[ConnectedApp]>` or `<VIMNullResponse>`, etc.

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -0,0 +1,244 @@
...
+
+    /// The type of the connected app, as a String.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `type` is preferred.
+    public private(set) var typeString: String?
+
+    /// The type of the connected app.
+    /// - Note: A returned value of `nil` indicates an unsupported `ConnectedAppType`, or a malformed response.
+    @nonobjc public var type: ConnectedAppType? {
+        switch self.typeString {
```
rpassis: What do you think of moving this to the `ConnectedAppType` instead? Does it feel weird to conform to RawRepresentable `String` when the backing type of the enum is `Int`?

```
extension ConnectedAppType: RawRepresentable {
    public typealias RawValue = String
    public var rawValue: RawValue { return String(describing: self) }
    public init?(rawValue: Self.RawValue) {
        switch rawValue { ... }
    }
}
```

Sources/Shared/Models/VIMInteraction.m
```diff
@@ -81,14 +93,18 @@ - (void)setStreamStatus {
...
-    
+
     NSAssert(number != nil, @"VOD video stream status not handled, unknown stream status");
-    
+
     self.streamStatus = [number intValue];
 }
 
+- (BOOL)isConnectedToApp {
+    return self.isConnected == TRUE;
```
rpassis: replace `TRUE` with `YES`

rpassis (approved):
@jasonhawkins Two comments for you to consider, and feel free to merge once they are addressed!

Sources/Shared/Models/ConnectedApp.swift
jasonhawkins: That's a good suggestion, thanks. I've change the `type var` implementation to this:

```Swift
@nonobjc public var type: ConnectedAppType? {
    typeString.map { ConnectedAppType.init(rawValue: $0) } ?? nil
}
```

So we can still return nil if something goes wrong, or we encounter an unsupported platform. 

But the backing type for `ConnectedAppType` is an Int only for Objective-C interoperability.

rpassis: Could probably go point free `typestring.map(ConnectedAppType.init)`

jasonhawkins: I'd still need to `?? nil` otherwise I'm trying to return an optional optional.

PR #391, opened 2019-12-02
VIM-7487: Adds support for publish jobs

jasonhawkins:
### Ticket
[VIM-7487](https://vimean.atlassian.net/browse/VIM-7487)

### Pull Request Checklist

### Ticket Summary
This PR builds on the recent [Connected Apps](https://github.com/vimeo/VimeoNetworking/pull/390) work by introducing the ability to create and transmit publish jobs to the Vimeo backend for distribution across multiple platforms simultaneously.

#### PublishDestination.swift
This file contains both a single `PublishDestination` class and a `PublishDestinations` class. The former represents a single publish destination, returned after a publish job has been sent to the server. The latter encapsulates all publish destinations for the supported platforms (Facebook, YouTube, etc.)

#### PublishJob.swift
A `PublishJob` object is returned by the server when the client has successfully requested to publish to one or more platforms. This object includes a unique `resourceKey`, date of first publish, as well as specific information related to each of the supported destinations.

#### PublishJobConnection.swift
This is a `VIMConnection` subclass that also includes data about blocking issues that will prevent a publish job from succeeding, known publishing constraints for each platform (duration, file size), and whether or not a destination has already been published to.

Because this connection adds properties to a `VIMConnection` type it cannot be parsed or accessed in the same way as a normal connection. That is, `video.connection(withName: <#ConnectionNameConstant#>)` will not work. Instead,  a `publishJobConnection` property for accessing this connection has been added to `VIMVideo`. If the connection is unavailable the property will be `nil`.

`PublishJobConnection.swift` also includes the implementation for the three new connection properties: `PublishJobBlockers`, `PublishJobConstraints`, and `PublishJobDestinations`.

#### SocialMediaPosts.swift
The client is responsible for creating social media post objects to send a video to the Vimeo server for distribution. The `SocialMediaPosts` class represents the data that will actually be sent to the server, and is composed of optional posts for each platform. A `SocialMediaPosts` object can be initialized with one or more posts, and can be mutated later on if needed (if, for example, a user creates a Facebook post but then later decides to also publish to YouTube).

#### VIMVideo
Updated to support the added `publishJobConnection` property, as well as its specific parsing requirements, which fall outside of how `VIMConnections` are typically parsed.

#### Request+PublishToSocial.swift
Added requests for sending `SocialMediaPosts` to the Vimeo server for distribution, and for fetching existing `PublishJobs`. 

Tests covering all of the above work have also been added.

### Reviewer Tips
Testing this work requires access to the internal branch, `https://pts-staging-api.ci.vimeows.com`. Furthermore, these features will only be available to whitelisted apps for the foreseeable future.

### How to Test
Ensure unit tests pass. If testing against the CI branch ensure your client application is able to make the included requests and parse the results.


cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 231 tests, with 0 failures (0 unexpected) in 1.118 (1.507) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1653)

</p>

antonklevets:
@jasonhawkins I receive compilation errors when try use public struct for requests.
Like
`'PublishToFacebookPost' initializer is inaccessible due to 'internal' protection level`
`'SocialMediaPosts' initializer is inaccessible due to 'internal' protection level`

All public struct need a public init.
https://stackoverflow.com/questions/54673224/public-struct-in-framework-init-is-inaccessible-due-to-internal-protection-lev

Sources/Shared/Models/VIMConnection.m
```diff
@@ -67,6 +67,7 @@
 NSString *const VIMConnectionNameBlockUser = @"block";
 NSString *const VIMConnectionNameLiveStats = @"live_stats";
 NSString *const VIMConnectionNameUploadAttempt = @"upload_attempt";
+NSString *const VIMConnectionNamePublishToSocialJob = @"publish_to_social_job";
```
antonklevets: Seems  should be `publish_to_social` (i.e. without "job" suffix). I checked via https://pts-staging.ci.vimeows.com/, there no `publish_to_social_job` for video objects, only `publish_to_social`

blueprint https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-4dafa2f9b34cb46c64cb8ef8d9694149R124

Sources/Shared/Models/VIMVideo.m
```diff
@@ -160,6 +161,11 @@ - (Class)getClassForObjectKey:(NSString *)key
         return [FileTransfer class];
     }
 
+    if ([key isEqualToString:@"publish_to_social_job"])
```
antonklevets: Seems should be `publish_to_social`

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,98 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+@objcMembers
+public class PublishJobConnection: VIMConnection {
+    public var publishBlockers: PublishJobBlockers?
+    public var publishConstraints: PublishJobConstraints?
+
```
antonklevets, line 40: Should add `publishDestinations` also 
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-4dafa2f9b34cb46c64cb8ef8d9694149R157

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,98 @@
...
+    public var linkedin: [String]?
+    public var twitter: [String]?
+}
+
+@objcMembers
+public class PublishConstraints: VIMModelObject {
+    public var duration: NSNumber?
+    public var size: NSNumber?
+}
+
```
antonklevets, line 92: Should add PublishDestiantions class for map `publish_destinations` object. like
```
public class PublishDestinations: VIMModelObject {
    public var facebook: Bool
    public var youtube: Bool
    public var linkedin: Bool
    public var twitter: Bool
}
```

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+    
+    struct Constraints {
+        static let size = "size"
+        static let duration = "duration"
+    }
+    
+    struct Key {
+        static let publishBlockers = "publish_blockers"
+        static let publishConstraints = "publish_constraints"
```
antonklevets: need add `publish_destinations` key

Sources/Shared/Models/PublishViewModels.swift
```diff
@@ -0,0 +1,175 @@
...
+public struct PublishToFacebookPost {
+
+    /// The title of the post as it will appear on Facebook.
+    public let title: String
+
+    /// The description of the post as it will appear on Facebook.
+    public let description: String
+
+    /// The identifier of the page being posted to on Facebook.
+    public let pageID: Int
```
antonklevets: instead of `pageID: Int` must be `destination: Int`
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-1f1bcce0884b841f7fd76d22ef138caeR27

Sources/Shared/Models/PublishViewModels.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    /// The Facebook cateogry of the video.
+    public let categoryID: Int
```
antonklevets: Should be `string`
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-1f1bcce0884b841f7fd76d22ef138caeR28

Sources/Shared/Models/PublishViewModels.swift
```diff
@@ -0,0 +1,175 @@
...
+    public let description: String
+
+    /// A list of tags for the video on YouTube. May be nil.
+    public let tags: [String]?
+
+    /// The privacy option for this video on YouTube.
+    public let privacy: Privacy
+
+    /// The YouTube category identifier to which this video falls into. May be nil.
+    public let categoryID: Int?
```
antonklevets: Should be `String?`
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-1f1bcce0884b841f7fd76d22ef138caeR47

Sources/Shared/Models/VIMVideo.h
```diff
@@ -41,6 +41,7 @@
 @class VIMUpload;
 @class FileTransfer;
 @class APIClientApp;
+@class PublishToSocialJob;
```
antonklevets: This class not exists. Seems shoulde be `PublishJobConnection`

Sources/Shared/Models/VIMVideo.h
```diff
@@ -96,6 +97,7 @@ typedef NS_ENUM(NSUInteger, VIMVideoProcessingStatus) {
 @property (nonatomic, strong, nullable) VIMUpload *upload;
 @property (nonatomic, strong, nullable) FileTransfer *fileTransfer;
 @property (nonatomic, strong, nullable) APIClientApp *sourceClientApp;
+@property (nonatomic, strong, nullable) PublishToSocialJob *publishToSocialJob;
```
antonklevets: This class not exists. Should be `PublishJobConnection *publishToSocial`

Sources/Shared/Models/VIMVideo.m
```diff
@@ -80,7 +80,8 @@ - (NSDictionary *)getObjectMapping {
         @"play": @"playRepresentation",
         @"review_page": @"reviewPage",
         @"file_transfer": @"fileTransfer",
-        @"app": @"sourceClientApp"
+        @"app": @"sourceClientApp",
+        @"publish_to_social": @"publishToSocialJob"
```
antonklevets: See below comment. If you will rename to `publishToSocial`, then this should be renamed to `publishToSocial` also

Sources/Shared/Models/VIMVideo.m
antonklevets: Should be `publish_to_social`
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-4dafa2f9b34cb46c64cb8ef8d9694149R124

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -0,0 +1,106 @@
...
+    ///   - videoID: The identifier for the video to be published.
+    static func publishPosts(_ posts: SocialMediaPosts, for videoID: String) -> Request {
+        let publishToSocialPath = String.videosURI + "/\(videoID)" + String.publishToSocialURI
+        var parameters = [String: Any]()
+
+        posts.facebook.map {
+            var post = [String: Any]()
+            post[String.Key.title] = $0.title
+            post[String.Key.description] = $0.description
+            post[String.Key.pageID] = $0.pageID
```
antonklevets: `post[String.Key.destination] = $0.destination`

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -0,0 +1,106 @@
...
+    static let publishToSocialURI = "/publish_to_social"
+
+    struct Key {
+        static let facebook = "facebook"
+        static let linkedin = "linkedin"
+        static let twitter = "twitter"
+        static let youtube = "youtube"
+        static let title = "title"
+        static let description = "description"
+        static let pageID = "page_id"
```
antonklevets: `page_id` used for linkedin. For facebook need add `destiantion` key
https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-1f1bcce0884b841f7fd76d22ef138caeR27

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -25,15 +25,18 @@
 
 @objcMembers
 public class PublishJobConnection: VIMConnection {
-    public var publishBlockers: PublishJobBlockers?
-    public var publishConstraints: PublishJobConstraints?
+    public private(set) var publishBlockers: PublishJobBlockers?
+    public private(set) var publishConstraints: PublishJobConstraints?
+    public private(set) var publishDestinations: PublishDestinations?
```
antonklevets: `PublishDestinations` class with multiple `PublishJobDestination` for each destination. But by blueprint and in response just `bool` values. 

https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-4dafa2f9b34cb46c64cb8ef8d9694149R157
```
  - `publish_destinations` (object) - Information on which social networks this video was published to.
     - `facebook` `true` (boolean, required) - Whether the video was ever published to Facebook.
     - `youtube` `true` (boolean, required) - Whether the video was ever published to YouTube.
     - `linkedin` `false` (boolean, required) - Whether the video was ever published to LinkedIn.
     - `twitter` `false` (boolean, required) - Whether the video was ever published to Twitter.
```

Seems wrong 😕

antonklevets:
@jasonhawkins Cannot get `PublishJobConnection` data from video object. 

I try `video?.connection(withName: "publish_to_social")` -> return `VIMConnection` object

When use `let connection = video?.connection(withName: "publish_to_social") as? PublishJobConnection`  -> connection always is nil

nicolelehrer:
@jasonhawkins just reading through description, so no rush to get back, but was wondering about `video.connection(withName: <#ConnectionNameConstant#>)` not working for a subclass of `VIMConnection`- don't we do that for `VIMVODConnection`?

jasonhawkins:
@nicolelehrer Yes, but the difference is that `VIMVODConnection` only has `NSNumber` properties. `PublishJobConnection` has properties which require their own mapping and parsing (`PublishJobBlockers`, `PublishJobConstraints`, `PublishJobDestinations`).

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,196 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+/// A 1VIMConnection1 subclass that includes additional data related to publishing to multiple social media platforms.
```
nicolelehrer: typo on 1VIMConnection1

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,196 @@
...
+            return nil
+        }
+    }
+}
+
+/// Reasons for which this video cannot be published, split by platform.
+@objcMembers
+public class PublishJobBlockers: VIMModelObject {
+
+    /// Current blockers that will prevent posting to Facebook. If `nil` publishign is not blocked for this platform.
```
nicolelehrer: typo on `publishign`

nicolelehrer: Comma after If `nil`, here and below

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -66,17 +67,29 @@ extension PublishStatus: CustomStringConvertible {
...
+    public var thirdPartyPostID: String?
+
+    /// The number of views this post has, as reported by the third party platform.
+    public var thirdPartyViewCount: NSNumber?
+
+    /// The number of likes (or equivalent) this post has, as reported by the third party platform.
+    public var thirdPartyLikeCount: NSNumber?
+
+    /// The number of comments this post has, as reported by the third party platform.
+    public var thirdPartyCommentCount: NSNumber?
```
antonklevets: In responses still doesn't exist these values. Waiting api updates for re-check it

Sources/Shared/Models/VIMVideo.m
```diff
@@ -287,6 +294,16 @@ - (void)parseConnections
     self.connections = connections;
 }
 
+- (void)parsePublishJobConnection
+{
+    NSDictionary *dictionary = [[self.metadata valueForKey:@"connections"] valueForKey:@"publish_to_social"];
+    if (dictionary != nil) {
+        VIMObjectMapper *mapper = [[VIMObjectMapper alloc] init];
+        [mapper addMappingClass: [PublishJobConnection class] forKeypath:@""];
+        self.publishJobConnection = [mapper applyMappingToJSON:dictionary];
```
nicolelehrer, line 303: Would it have been possible to include this in `parseConnections`, but create a helper that maps for the special case if the key is equal to `publish_to_social` (when given a dictionary, it outputs a connection); then that connection would be added to all of the connections via ` [connections setObject:connection forKey:key];` so we could access it as we do other connections?

Sources/Shared/Models/VIMVideo.m
```diff
@@ -287,6 +294,16 @@ - (void)parseConnections
     self.connections = connections;
 }
 
+- (void)parsePublishJobConnection
+{
+    NSDictionary *dictionary = [[self.metadata valueForKey:@"connections"] valueForKey:@"publish_to_social"];
+    if (dictionary != nil) {
+        VIMObjectMapper *mapper = [[VIMObjectMapper alloc] init];
+        [mapper addMappingClass: [PublishJobConnection class] forKeypath:@""];
```
nicolelehrer, line 302: Should the keyPath be some top level object like `data` or something else?

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,196 @@
...
+public class PublishConstraints: VIMModelObject {
+
+    /// The maximum time in seconds for a video to be uploaded to a platform.
+    public private(set) var duration: NSNumber?
+
+    /// The maximum file size in bytes for a video to be uploaded to a platform.
+    public private(set) var size: NSNumber?
+}
+
+private extension String {
```
nicolelehrer, line 215: Since folks could still add a non-string into one of the nested structs, any advantage or need for this to extend string?  Should this just be a constants struct?

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Whether the video was ever published to Twitter.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `twitter` is preferred.
+    public private(set) var publishedToTwitter: NSNumber?
+
+    /// Whether the video was ever published to YouTube.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `youtube` is preferred.
+    public private(set) var publishedToYouTube: NSNumber?
```
nicolelehrer: Could this not be mapped internally as an `NSNumber` and then made a public accessor as a `bool`? Can we not do that as the mapper then won't see these? If that's the case could we make these internal maybe?

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -0,0 +1,196 @@
...
+
+/// Reasons for which this video cannot be published, split by platform.
+@objcMembers
+public class PublishJobBlockers: VIMModelObject {
+
+    /// Current blockers that will prevent posting to Facebook. If `nil` publishign is not blocked for this platform.
+    public private(set) var facebook: [String]?
+
+    /// Current blockers that will prevent posting to YouTube. If `nil` publishing is not blocked for this platform.
+    public private(set) var youtube: [String]?
```
nicolelehrer: Do we ever care about the non-nil values?  If so are there expected values / if not should we just create api like `facebookIsBlocked: Boolean` etc

Sources/Shared/VimeoSessionManager.swift
```diff
@@ -41,7 +41,7 @@ final public class VimeoSessionManager: NSObject, SessionManaging {
     // MARK: - Public
 
     /// Base URL for the Vimeo API
-    public internal(set) static var baseURL = URL(string: "https://api.vimeo.com")!
+    public internal(set) static var baseURL = URL(string: "https://pts-staging-api.ci.vimeows.com")!
```
nicolelehrer: Should this be reverted for the PR now, and just ask testers to update it manually so we don't accidentally forget to revert?

Sources/Shared/Models/PublishJob.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+
+/// Data related to all supported publishing platforms, including date of first publish and a unique resource key.
+@objcMembers
+public class PublishJob: VIMModelObject {
+    
+    /// The time in ISO 8601 format when the user first published this post.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `firstPublishDate` is preferred.
+    public private(set) var firstPublishDateString: String?
```
nicolelehrer: Can any of the objc-c interop properties be hidden from use with access control by the client? would `internal` do it without hindering mapping?

Sources/Shared/Models/PublishJob.swift
```diff
@@ -0,0 +1,88 @@
...
+private extension String {
+    struct Key {
+        static let destinations = "destinations"
+        static let firstPublishDate = "first_publish_date"
+        static let resourceKey = "resource_key"
+    }
+    
+    struct Value {
+        static let firstPublishDate = "firstPublishDateString"
+        static let resourceKey = "resourceKey"
```
nicolelehrer: Also wondering here if nesting should be used instead of just labelling the static props with `...Key` and `...Value` at the end so we can ensure string type safety? Or wrapping these all in a constants struct instead of in String extension?

Sources/Shared/Models/PublishJob.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// The time in ISO 8601 format when the user first published this post.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `firstPublishDate` is preferred.
+    public private(set) var firstPublishDateString: String?
+    
+    /// The time as a `Date` when the user first published this post.
+    public private(set) var firstPublishDate: Date?
+    
+    /// The status of the post on each of the supported platforms.
+    public private(set) var destinations: PublishDestinations?
```
nicolelehrer, line 36: (This is a nit pick, can't really think of better alternatives, so we can punt on this but) - It's tough, because this makes me think of an array of destinations when really it's a class that holds all of them. I know everyone here hates the word `container` ... can't think of any good alternatives - but overall wondering if there's a name that can be in the singular but would reflect an object that encapsulate all social destinations.

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+
+extension PublishStatus: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case .error:
+            return String.error
+        case .finished:
+            return String.finished
+        case .inProgress:
+            return String.inProgress
```
nicolelehrer: can we remove each `String.` here?

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+public class PublishDestination: VIMModelObject {
+    /// The status of the post on the specified platform as a `String`.
+    /// - Note: This property is available to provide interoperability with Objective-C codebases.
+    ///         Using `status` is preferred.
+    public private(set) var statusString: String?
+    
+    /// The status of the post for a given platform.
+    @nonobjc public var status: PublishStatus? {
+        switch self.statusString {
+        case String.error:
```
nicolelehrer, line 56: Can this not find `.error` on String (that is without using the `String.`) because `statusString` is optional?

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+    /// The URL of the post on a given platform.
+    public private(set) var thirdPartyPostURL: String?
+    
+    /// The ID of the post on a given platform.
+    public private(set) var thirdPartyPostID: String?
+
+    /// The number of views this post has, as reported by the third party platform.
+    public private(set) var thirdPartyViewCount: NSNumber?
+
+    /// The number of likes (or equivalent) this post has, as reported by the third party platform.
```
nicolelehrer, line 75: What does `or equivalent` mean in this context?

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+}
+
+private extension String {
+    struct Key {
+        static let facebook = "facebook"
+        static let linkedin = "linkedin"
+        static let twitter = "twitter"
+        static let youtube = "youtube"
```
nicolelehrer: Feel like I've seen some of these defined in multiple places, should we make a shared struct or define a more accessible extension on `String`?

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+    
+    // MARK: - Overrides
+    
+    public override func getObjectMapping() -> Any? {
+        return [
+            String.Key.thirdPartyPostURL: String.Value.thirdPartyPostURL,
+            String.Key.thirdPartyPostID: String.Value.thirdPartyPostID,
+            String.Key.thirdPartyViewCount: String.Value.thirdPartyViewCount,
+            String.Key.thirdPartyLikeCount: String.Value.thirdPartyLikeCount,
+            String.Key.thirdPartyCommentCount: String.Value.thirdPartyCommentCount,
```
nicolelehrer: I thought we get some of this underscore to camelCase functionality for free with object mapper - I mean it might be better to be explicit moving forward anyway, since when we deprecate it, it will be hard to know when that was used

Sources/Shared/Models/PublishDestination.swift
```diff
@@ -0,0 +1,153 @@
...
+/// - finished: The post was successfully published.
+/// - inProgress: The post creation process is currently underway.
+@objc public enum PublishStatus: Int {
+    case error
+    case finished
+    case inProgress
+}
+
+extension PublishStatus: CustomStringConvertible {
+    public var description: String {
```
nicolelehrer, line 37: `asString`?

Sources/Shared/Models/SocialMediaPosts.swift
```diff
@@ -0,0 +1,175 @@
...
+    /// The title of the post as it will appear on Facebook.
+    public let title: String
+
+    /// The description of the post as it will appear on Facebook.
+    public let description: String
+
+    /// The identifier of the Facebook page being posted to.
+    public let destination: Int
+
+    /// The Facebook cateogry of the video.
```
nicolelehrer: nit - typo on category

Sources/Shared/Models/SocialMediaPosts.swift
```diff
@@ -0,0 +1,175 @@
...
+    public let title: String
+
+    /// The description of the video as it will appear on YouTube.
+    public let description: String
+
+    /// A list of tags for the video on YouTube. May be nil.
+    public let tags: [String]?
+
+    /// The privacy option for this video on YouTube.
+    public let privacy: Privacy
```
nicolelehrer, line 129: Unless we think there could be more privacy options down the line could maybe use `isPrivate` bool flag to simplify

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -0,0 +1,107 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+public extension Request where ModelType == PublishJob {
+    /// Returns a request for fetching the `PublishJob` for a specified video ID.
+    /// - Parameter videoID: The identifier for the video to be published.
+    static func fetchPublishJob(for videoID: String) -> Request {
+        let publishToSocialPath = String.videosURI + "/\(videoID)" + String.publishToSocialURI
```
nicolelehrer: should this be broken out into a helper that builds path given `videoID` since it's used below as well

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -0,0 +1,107 @@
...
+            var post = [String: Any]()
+            post[String.Key.title] = $0.title
+            post[String.Key.description] = $0.description
+            post[String.Key.destination] = $0.destination
+            post[String.Key.categoryID] = $0.categoryID
+            post[String.Key.allowEmbedding] = $0.allowEmbedding
+            post[String.Key.shouldAppearOnNewsFeed] = $0.shouldAppearOnNewsFeed
+            post[String.Key.isSecretVideo] = $0.isSecretVideo
+            post[String.Key.allowSocialActions] = $0.allowSocialActions
+            parameters[String.Key.facebook] = post
```
nicolelehrer: Can post just be a constant that gets instantiated in one go with a literal dictionary? like
```swift
 let post = [String.Key.title: $0.title,
                   String.Key.categoryID: $0.categoryID] 
```

nicolelehrer (commented):
hey @jasonhawkins my review is still in progress, this is just a first pass and haven't reviewed tests yet, but releasing in case helpful - feel free to wait for full review if easier.

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: @antonklevets I added these based on the updated API spec, but you're right that these haven't been implemented on the API yet. It sounds like they may be available by end of day today, but it might be worth checking back tomorrow.

Tests/Fixtures/video-with-connection.json
```diff
@@ -0,0 +1,47 @@
+{
```
nicolelehrer: nit - should we make this a destination for multiple connections and rename to `connections` or make this specific to `publish-to-social-connection`

Sources/Shared/Requests/Request+PublishToSocial.swift
nicolelehrer: Any downside of injecting the `videoURI` itself instead of constructing it from the `videoID`

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -0,0 +1,107 @@
...
+        return Request(path: publishToSocialPath)
+    }
+
+    /// Returns a `put` request for publishing `SocialMediaPosts` for the given video ID.
+    /// - Note: Post data should be validated prior to creating the publish request. Posts that fail server-side
+    ///         validation will produce an error for that platform.
+    /// - Parameters:
+    ///   - posts: A structure containing all of the data necessary to publish to multiple platforms simultaneously.
+    ///   - videoID: The identifier for the video to be published.
+    static func publishPosts(_ posts: SocialMediaPosts, for videoID: String) -> Request {
```
nicolelehrer: Is this considered a type `Request<PublishJob>` since it's located in this extension, or do we expect something like `Request<VIMNullResponse>`

Tests/Shared/PublishToSocialPostsTests.swift
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        XCTAssertEqual(linkedInPost.pageID, 1234)
+        XCTAssertEqual(linkedInPost.title, "Important Information")
+        XCTAssertEqual(linkedInPost.description, "You are facing the wrong way.")
+    }
+
+    func test_publishToTwitterPost_canBeCreatedSuccessfully() {
+        let twitterPost = PublishToTwitterPost(
+            tweet: "just salad walk for lunch holla"
```
nicolelehrer, line 66: 😄

Tests/Shared/PublishToSocialPostsTests.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        XCTAssertEqual(youTubePost.title, "How to Publish Everywhere With a Single Click")
+        XCTAssertEqual(youTubePost.description, "Learn everything you need to know right here.")
+        XCTAssertEqual(youTubePost.tags?.count, 3)
+        XCTAssertEqual(youTubePost.tags, ["Vimeo", "Publish", "Educational"])
+        XCTAssertEqual(youTubePost.privacy, .public)
+        XCTAssertEqual(youTubePost.categoryID, "Space")
+    }
+
+    func test_socailMediaPostsObject_canBeCreatedSuccessfully() {
```
nicolelehrer: nit typo on social

nicolelehrer (changes requested):

nicolelehrer:
@jasonhawkins overall looks great, commented on mostly cosmetics, & tests pass 💯

Sources/Shared/Models/VIMVideo.m
jasonhawkins: I tried something like this originally, but ran into an issue when trying to cast the returned connection in Swift. `connection(withName:)` will still always return a `VIMConnection`, even if we try to cast it as `PublishJobConnection`, because the parent class doesn't know anything about its subclass.

But it sounds like what you're suggesting is that these new properties will just be nil for any other `VIMConnection`. Is that right? Would that mean adding those properties to `VIMConnection` directly, instead of creating the `PublishJobConnection` subclass?

Sources/Shared/Models/VIMVideo.m
jasonhawkins: The `publish_to_social` connection doesn't have a `data` object to worry about here. I captured a response from `/videos/277485394?fields=metadata.connections.publish_to_social` and created a test to validate that parsing worked as expected, which can be found at `PublishJobConnectionTests.swift:91`.

Sources/Shared/Models/PublishJobConnection.swift
jasonhawkins: I could, yes, but I was trying to follow the pattern where we extend foundation types as necessary. Is it preferred to just use constant structs instead?

Sources/Shared/Models/PublishJobConnection.swift
jasonhawkins: Do you mean changing this to `internal private(set)`? I like that, and these values don't necessarily need to be accessible at all anyway. I'll update the documentation to remove the lines about Objective-C interoperability to..

Sources/Shared/Models/PublishJobConnection.swift
jasonhawkins: Good point. We do care about the values returned in these arrays. I've refactored this to hide the arrays and instead return objects that can be inspected to determine whether or not a blocker was returned. Let me know what you think.

//cc @antonklevets This change may affect you if you've implemented blockers in your code client side already.

Sources/Shared/VimeoSessionManager.swift
jasonhawkins: Yes, absolutely

Sources/Shared/Models/PublishJob.swift
jasonhawkins: `internal` should work fine here, good idea.

Sources/Shared/Models/PublishJob.swift
jasonhawkins: I know what you mean, I think trying to figure out the naming for this stuff (and trying to distinguish between things like `PublishJobDestination` and `PublishDestination`) has been one of the most difficult parts to get a handle on.

In this case I went down the same path as you but couldn't think of anything better. I decided to leave it so that it matched what the API returned. But I'm happy to change it if we come up with anything better.

Sources/Shared/Models/VIMVideo.m
nicolelehrer: Ah ok, I see what you mean about the casting issue. Makes sense that you need to keep this separate. 

For the second part, I guess that is an approach we could take? But I think a unique connection object makes more sense at this point. 

If other consumers like Anton and Liron feel this is too divergent from the existing API then I would understand trying to keep the existing format but I'm not aware of any use cases where we need to iterate through all connections and do something with all of them. 

I'm down to keep as is barring cases mentioned above.

Sources/Shared/Models/PublishJobConnection.swift
nicolelehrer: My only thought here was that the advantage of extending the foundation types is the dot notation without needing to preface with `String`, but we lose that here with the nested struct. I would opt for making a generic constants struct or removing the nested ones, but happy to get another opinion on this one!

Sources/Shared/Models/PublishJob.swift
nicolelehrer: Ok to keep as is, thanks!

jasonhawkins: For reference, this is where it's defined in the blueprint: https://github.vimeows.com/Vimeo-API/design/pull/148/files#diff-be5517da643539b403699fe2238094b7R3

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: That seems to be the case, yeah.

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: I didn't want to paint all the different ways these companies represent positive engagement as a "like". Didn't Twitter used to use "favorites"? And who even knows what LinkedIn uses.

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: That seems pretty reasonable. It looks like I'll still have to include the `String` type when accessing these in most cases though.

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: I think I'd rather be explicit here, especially if an object mapping shakeup is coming any time soon.

Sources/Shared/Models/PublishDestination.swift
jasonhawkins: `description` is declared as part of the `CustomStringConvertible` protocol. It's documentation just says, "A textual representation of this instance."

Sources/Shared/Models/SocialMediaPosts.swift
jasonhawkins: Good question. I checked with API and apparently there's a third option for `unlisted`, but it's not yet supported on the web front end. But it could at some point.

Sources/Shared/Requests/Request+PublishToSocial.swift
jasonhawkins: Yes, except for the optional parameters. I'll update and clean this stuff up though.

Tests/Fixtures/video-with-connection.json
jasonhawkins: Good call, I'm going to keep it specialized and call it `video-with-publish-to-social-connection.json`

Tests/Shared/PublishToSocialPostsTests.swift
jasonhawkins: 🥗🚶🏽‍♀️🕛 📣

Sources/Shared/Requests/Request+PublishToSocial.swift
jasonhawkins: This request will in fact return a `PublishJob` when it succeeds, including the date it was published and the current status of the job.

Sources/Shared/Models/PublishJobConnection.swift
antonklevets: I prefer this code. There can be: `nil` or `[size]` or `[size, duration]`, i.e. it's more flexible

Sources/Shared/Models/SocialMediaPosts.swift
nicolelehrer: Cool makes sense to keep as is

Sources/Shared/Models/PublishJobConnection.swift
jasonhawkins: That makes sense, yeah. 

I made some changes and put the strings that are only used as part of a Key / Value into a `Constants` struct, which will also keep the `Key` and `Value` structs. My reasoning is that I want to make it clear what those strings are for, and to be able to use the same name for these pairs.

The second part is that I moved all of the other strings into actual private String extensions. This lets me omit the type in _most_ cases, but the compiler still complained in others. For example, when creating the LinkedIn post the compiler couldn't figure out what these values meant without prefixing the `String` type:

```Swift
posts.linkedIn.map {
    parameters[.linkedin] = [
        String.pageID: $0.pageID,
        String.title: $0.title,
        String.description: $0.description,
    ]
}
```

Sources/Shared/Models/PublishJob.swift
jasonhawkins: In this case `resourceKey` is just the name of the property we're mapping.

jasonhawkins:
@nicolelehrer Thanks for the feedback on this, Nicole. It's much appreciated. This is ready for another look 👀

Sources/Shared/Models/ConnectedApp.swift
```diff
@@ -182,20 +182,20 @@ public class ConnectedApp: VIMModelObject {
...
-            String.Key.publishCategories: String.Value.publishCategories,
-            String.Key.thirdPartyUserID: String.Value.thirdPartyUserID,
-            String.Key.thirdPartyUserDisplayName: String.Value.thirdPartyUserDisplayName,
-            String.Key.type: String.Value.type
+            Constants.Key.addDate: Constants.Value.addDate,
+            Constants.Key.dataAccessIsExpired: Constants.Value.dataAccessIsExpired,
+            Constants.Key.publishCategories: Constants.Value.publishCategories,
+            Constants.Key.thirdPartyUserID: Constants.Value.thirdPartyUserID,
+            Constants.Key.thirdPartyUserDisplayName: Constants.Value.thirdPartyUserDisplayName,
+            Constants.Key.type: Constants.Value.type
```
nicolelehrer, line 190: @jasonhawkins what is we used tuples to store the constants for the key-value pairs so we could facilitate the correct pairing? For example

```swift
static let addDate = (key: "add_date", value: "addDateString")
```
would mean we could do 

```swift
[Constants.addDate.key: Constants.addDate.value]
```
And if you left of the key words you could make it even shorter as 

```swift
[Constants.addDate.0: Constants.addDate.1]
```

Sources/Shared/Models/SocialMediaPosts.swift
```diff
@@ -35,8 +35,8 @@ public struct PublishToFacebookPost {
     /// The identifier of the Facebook page being posted to.
     public let destination: Int
 
-    /// The Facebook cateogry of the video.
-    public let categoryID: String
+    /// The Facebook category of the video.
+    public let categoryID: String?
```
nicolelehrer, line 39: nit that we point out elsewhere when params can be nil

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -36,72 +36,73 @@ public extension Request where ModelType == PublishJob {
...
+            $0.description.map { (description) in post[.description] = description }
+            $0.tags.map { (tags) in post[.tags] = tags }
+
+            parameters[.youtube] = post
         }
 
         return Request(method: .put, path: publishToSocialPath, parameters: parameters)
     }
+
+    private static func uri(for videoURI: String) -> String {
```
nicolelehrer: Could see this being named `publishToSocialPath(for videoURI: String)` or just `path(for...`

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -36,72 +36,73 @@ public extension Request where ModelType == PublishJob {
...
+                .title: $0.title,
+                .description: $0.description,
+                .destination: $0.destination,
+                .allowEmbedding: $0.allowEmbedding,
+                .shouldAppearOnNewsFeed: $0.shouldAppearOnNewsFeed,
+                .isSecretVideo: $0.isSecretVideo,
+                .allowSocialActions: $0.allowSocialActions
+            ]
+            
+            $0.categoryID.map { (categoryID) in post[.categoryID] = categoryID }
```
nicolelehrer, line 55: Could this be `$0.categoryID.map { post[.categoryID] = $0 }`? perhaps reasonable for devs to differentiate from other $0 since we're in new closure scope? if not ok to keep as is

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -36,72 +36,73 @@ public extension Request where ModelType == PublishJob {
...
         posts.linkedIn.map {
-            var post = [String: Any]()
-            post[String.Key.pageID] = $0.pageID
-            post[String.Key.title] = $0.title
-            post[String.Key.description] = $0.description
-            parameters[String.Key.linkedin] = post
+            parameters[.linkedin] = [
+                String.pageID: $0.pageID,
+                String.title: $0.title,
+                String.description: $0.description,
```
nicolelehrer: Extra comma on end here

nicolelehrer: Getting a warning here `Expression implicitly coerced from 'String?' to 'Any'` 
- Should description be optional? If so, do we want to provide a default value or will it get removed if it's nil?
- Otherwise should we cast as `Any`

Sources/Shared/Requests/Request+PublishToSocial.swift
```diff
@@ -36,72 +36,73 @@ public extension Request where ModelType == PublishJob {
...
-            $0.tags.map { (tags) in  post[String.Key.tags] = tags }
-            parameters[String.Key.youtube] = post
+            var post: [String: Any] = [
+                .title: $0.title,
+                .privacy: $0.privacy.rawValue,
+                .categoryID: $0.categoryID
+            ]
+
+            $0.description.map { (description) in post[.description] = description }
+            $0.tags.map { (tags) in post[.tags] = tags }
```
nicolelehrer, line 83: Since `tags` is an optional array, how does this map? Does it pass the entire array or does it only take the last value?

nicolelehrer (changes requested):

Sources/Shared/Models/ConnectedApp.swift
jasonhawkins: Ohh nice. It'll work for things that are 1:1 key-value pairs, like mapping properties in an object. But for cases where we have a key that maps to a class, the value is derived from the name of the class and not an explicitly defined String: 

```Swift
public override func getClassForObjectKey(_ key: String?) -> AnyClass? {
    switch key {
    case Constants.Key.publishBlockers:
        return PublishJobBlockers.self
    default:
        return nil
    }
}
```

I'll use this approach for the 1:1 pairs, and move the orphans into the private String extensions.

Sources/Shared/Requests/Request+PublishToSocial.swift
jasonhawkins: Reasonable, perhaps yeah. But also probably a slightly better chance of misreading it at a glance. I think I'd prefer to keep it as-is if that's cool.

Sources/Shared/Requests/Request+PublishToSocial.swift
jasonhawkins: Yeah, this was a mistake. It should be fixed now.

Sources/Shared/Requests/Request+PublishToSocial.swift
jasonhawkins: Good question, this will assign the entire array. This should be validated in `Request+PublishToSocialTests.swift:158`:

```Swift
let socialMediaPosts = SocialMediaPosts(youTube: youTubePost)
let request = Request.publishPosts(socialMediaPosts, for: "/videos/34567")
let parameters = try XCTUnwrap(request.parameters as? [String: Any])
let youtubeParameters = try XCTUnwrap(parameters["youtube"] as? [String: AnyHashable])
…
XCTAssertEqual(youtubeParameters["tags"], ["test", "tags"])
```

jasonhawkins:
Ready again, @nicolelehrer

nicolelehrer (approved):
🥗 🚶 🎉

PR #393, opened 2019-12-11
MP-40: Add video transcode status

nicolelehrer:
#### Ticket

[MP-40](https://vimean.atlassian.net/browse/MP-40)

#### Pull Request Checklist

#### Issue Summary

- Adds a video transcode status model and request builder 

#### Implementation Summary

This PR uses the new Networking API compatible with codable models
- `TranscodeStatusEndpointType`: defines the request to `/videoURI/status` 
- `VideoTranscodeStatus`: decodable type defining the video status object 
  - [API docs for Transcode status](https://developer.vimeo.com/api/reference/videos#get_transcode_status) - note, must be logged in as a Staff Account to see docs 
  - Basis for `TranscodeState` enum values: [link to backend code](https://github.vimeows.com/Vimeo/vimeo/blob/df07e4b924275827fc68e92c733d2ae795667b02/src/Vimeo/Api/Response/ClipStatusResponse.php#L51)
- `VideoTranscodeStatusTests`: tests decoder since we defined it ourselves in order to map to an enum type

#### Reviewer Tips

- Because this is a self contained request, it seemed like a safe place to use out new networking API 

- How would this code be used by a client? Pass your custom endpoint type into this method to make a request:
```swift
public func request<Model: Decodable>(
        _ endpoint: EndpointType,
        startImmediately: Bool = true,
        completionQueue: DispatchQueue = .main,
        then callback: @escaping (Result<Model, Error>) -> Void
    ) -> RequestToken 
``` 

For example 

```swift
let endpointType = TranscodeStatusEndpointType(videoURI: "videos/xxxxxx")

_ = VimeoClient.shared.request(endpointType) { (result: Result<VideoTranscodeStatus, Error>) in
   switch result {
      case .success(let result):
        // handle result
      case .failure(let error):
        // handle error
   }
}
```

#### How to Test

- Confirm added tests pass
- To test fetching data from the API 
  - find a video you own and get its uri
  - log into the account owning the video
  - run the example code provided above with the uri of the video you own
  - you should see that you get a valid `VideoTranscodeStatus` object back, such as `VideoTranscodeStatus(state: VimeoNetworking.VideoTranscodeStatus.TranscodeState.ready, progress: 100, timeLeft: 0)` for a video that is already transcoded and ready watch


Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
```diff
@@ -0,0 +1,49 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+/// Encapsulates all information required to fetch a video's transcode status
+public struct TranscodeStatusEndpointType: EndpointType {
```
rpassis: I'd normally just call this `TranscodeStatusEndpoint`

Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
```diff
@@ -0,0 +1,49 @@
...
+    public let headers: HTTPHeaders? = nil
+    public let parameters: Any? = nil
+    public let method = HTTPMethod.get
```
rpassis: `.get` should suffice with implicit member expression

Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
```diff
@@ -0,0 +1,49 @@
...
+    public let path: String
+    private let videoURI: String
+    private let statusPathComponent = "/status"
+    
+    /// Initializer for creating a TranscodeStatusEndpointType
+    /// - Parameter videoURI: the URI of the video
+    public init(videoURI: String) {
+        self.videoURI = videoURI
```
rpassis: why do we need to store the videoURI in this request? It only seems to be used by the initializer for create the `path`?

Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.path = videoURI + statusPathComponent
+    }
+    
+    /// Provides a configured request for fetching a video's transcode status
+    public func asURLRequest() throws -> URLRequest {
+        return URLRequest(url: self.baseURL.appendingPathComponent(path))
```
rpassis: `self.path` (unless we no longer care for explicit self, which I am all for!)

Sources/Shared/Models/VideoTranscodeStatus.swift
```diff
@@ -0,0 +1,90 @@
...
+public struct VideoTranscodeStatus: Decodable {
+    
+    /// Current state of transcoding
+    var state: TranscodeState
+    
+    /// Percentage of transcoding completed
+    var progress: Int
+    
+    /// Time in seconds remaining until completion
+    var timeLeft: Double
```
rpassis: if this is in seconds should it just be an `int`?

Sources/Shared/Models/VideoTranscodeStatus.swift
```diff
@@ -0,0 +1,90 @@
...
+        case ready
+        case retrieved
+        case standby
+        case starting
+        case unknown
+        case uploadComplete = "upload_complete"
+        case uploadIncomplete = "upload_incomplete"
+    }
+    
+    public init(from decoder: Decoder) throws {
```
rpassis: do you actually need this here? It seems all properties are decodable so the compiler should be able to synthesize this for us?

Tests/Shared/VideoTranscodeStatusTests.swift
```diff
@@ -0,0 +1,96 @@
...
+           XCTAssertNotNil(context)
+       }
+       catch {
+           XCTFail("Unexpected failure")
+       }
+    }
+    
+    // MARK: - TranscodeStatusEndpointType Tests
+
+    func testTranscodeStatusEndpointType_returnsCorrectURL() {
```
rpassis: I know this may seem like overkill but how about moving this to a separate testfile since it is testing a different component? This will make it easier when we move models and their tests out of this repo.

Tests/Shared/VideoTranscodeStatusTests.swift
```diff
@@ -0,0 +1,96 @@
...
+       }
+       catch let DecodingError.dataCorrupted(context) {
+           XCTAssertNotNil(context)
+       }
+       catch {
+           XCTFail("Unexpected failure")
+       }
+    }
+    
+    func testCustomDecoder_returnsTypeMismatchError_onInvalidTimeLeft_valueEmptyString() {
```
rpassis: You can avoid all these do/catch blocks by marking your test function as throwing i.e.

```
func testThatThisHappens() throws {
   try myObject.somethingThatThrows()
}
```

This would mean that we are not asserting on the `DecodingError`s but that is probably unnecessary since they are not implemented by us.

rpassis (changes requested):

Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
nicolelehrer: Ah, I thought so too, but compiler says `Reference to member 'get' cannot be resolved without a contextual type`

Sources/Shared/EndpointTypes/TranscodeStatusEndpointType.swift
nicolelehrer: yup no need good catch

Sources/Shared/Models/VideoTranscodeStatus.swift
nicolelehrer: When measuring background upload duration to report to fresnel, I have encountered issues client side where time goes on forever and returns a huge value, causing overflow. I assumed it might be best to safeguard against this case if transcoding happens to hang for a very long time.

Sources/Shared/Models/VideoTranscodeStatus.swift
nicolelehrer: Yes! Thank you for catching this! I'm deleting all the related tests

Sources/Shared/Models/VideoTranscodeStatus.swift
rpassis: `Int.max` in 64bits will be `18446744073709551615` seconds - do you think it is worth it making this into a `Double` just for that edge case? Then we have to deal with all the difficulties associated with that type i.e. I can see us having to cast this to `Int` to use with `TimeInterval`, comparisons not being possible etc.

I don't feel super strongly about it but from a consumer of this type I'd definitely prefer it to be an `Int` instead.

nicolelehrer: Given we won't crash at run time for this model, and can just catch an error, it's not worth using a sub-optimal type for edge cases. Fair point, down to update to `Int`. 

Just checking for my knowledge, Isn't `TimeInterval` typealias for `Double`?

Tests/Shared/VideoTranscodeStatusTests.swift
nicolelehrer: Deleting these but great call out, def the better way to go.

rpassis (approved):
LGTM 👍

Sources/Shared/Models/VideoTranscodeStatus.swift
rpassis: Wow who knew I always thought it was an `Int`! Given that I think it is actually fine to leave it as a double if you want to revert it, sorry about that!

cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 180 tests, with 0 failures (0 unexpected) in 1.237 (1.558) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1545)

</p>

rpassis (approved):

PR #403, opened 2020-01-24
Adds additional YouTube-specific publishing blocker

jasonhawkins:
### Pull Request Checklist

### Implementation Summary
Added `yt_no_channel` as a blocker for publishing to YouTube, when the destination account doesn't yet have any channels to publish to. 

This can be accessed in the same way that other blockers are accessed: `connection.publishBlockers?.youtube?.noChannels`

### How to Test
Ensure that unit tests pass.

### DIFF
--- a/Sources/Shared/Models/PublishJobConnection.swift
+++ b/Sources/Shared/Models/PublishJobConnection.swift
@@ -76,8 +76,8 @@ public class PublishJobBlockers: VIMModelObject {
     }()
 
     /// Current blockers that will prevent posting to YouTube. If `nil`, publishing is not blocked for this platform.
-    public lazy var youtube: PublishBlockers? = {
-        return PublishBlockers(blockers: self.youtubeBlockers)
+    public lazy var youtube: YouTubeBlockers? = {
+        return YouTubeBlockers(blockers: self.youtubeBlockers)
     }()
 
     public override func getObjectMapping() -> Any! {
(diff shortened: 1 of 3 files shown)


antonklevets (approved):

Sources/Shared/Models/PublishJobConnection.swift
```diff
@@ -76,8 +76,8 @@ public class PublishJobBlockers: VIMModelObject {
     }()
 
     /// Current blockers that will prevent posting to YouTube. If `nil`, publishing is not blocked for this platform.
-    public lazy var youtube: PublishBlockers? = {
-        return PublishBlockers(blockers: self.youtubeBlockers)
+    public lazy var youtube: YouTubeBlockers? = {
```
rpassis, line 79: is this the right name for this variable? I don't have the full context but perhaps it could be more explicit?

rpassis: Actually nevermind I just expanded the code above this change and noticed it is namespaced inside `PublishJobBlockers`

rpassis (approved):

jasonhawkins:
@rpassis Thanks, Anton and Rog 👍

Sources/Shared/Models/PublishJobConnection.swift
jasonhawkins: Yeah, I tried to keep the naming as similar to the API as I could. But thanks for raising it.

PR #407, opened 2020-02-07
MPS-270: Fix Crash Related to Comments Count

supperspidey:
## Ticket

[MPS-270](https://vimean.atlassian.net/browse/MPS-270)

## Pull Request Checklist

## Issue Summary

This PR fixed the bug that caused a method call to `commentsCount()` of `VIMVideo` to crash the consumer apps. Now, the implementation of this method first checks if the `total` property of `VIMConnection` &mdash; which is of `NSNumber` type &mdash; can respond to the `intValue` property before attempting to get the value out.

## Implementation Summary

N/A

## Reviewer Tips

N/A

## How to Test

N/A

### DIFF
--- a/Sources/Shared/Models/VIMConnection.m
+++ b/Sources/Shared/Models/VIMConnection.m
@@ -86,7 +86,7 @@ NSString *const VIMConnectionNamePublishToSocial = @"publish_to_social";
 - (void)didFinishMapping
 {
     if (self.total != nil && [self.total isKindOfClass: [NSNumber class]] == NO) {
-        NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
+        NSAssert(false, @"Error: Detected instance where `connection.total` is unexpectedly not an `NSNumber`.");
         self.total = @0;
     }
 }
--- a/Sources/Shared/Models/VIMConnection.m
+++ b/Sources/Shared/Models/VIMConnection.m
@@ -85,7 +85,7 @@ NSString *const VIMConnectionNamePublishToSocial = @"publish_to_social";
 
 - (void)didFinishMapping
 {
-    if ([self.total isKindOfClass: [NSNumber class]] == NO) {
+    if (self.total != nil && [self.total isKindOfClass: [NSNumber class]] == NO) {
         NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
         self.total = @0;
     }
--- a/Sources/Shared/Models/VIMConnection.m
+++ b/Sources/Shared/Models/VIMConnection.m
@@ -83,4 +83,12 @@ NSString *const VIMConnectionNamePublishToSocial = @"publish_to_social";
     return (self.options && [self.options containsObject:@"POST"]);
 }
 
+- (void)didFinishMapping
+{
+    if ([self.total isKindOfClass: [NSNumber class]] == NO) {
+        NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
+        self.total = @0;
+    }
+}
+
 @end
--- a/Sources/Shared/Models/VIMVideo.m
+++ b/Sources/Shared/Models/VIMVideo.m
@@ -542,21 +542,17 @@ NSString *VIMContentRating_Safe = @"safe";
 {
     VIMConnection *commentsConnection = [self connectionWithName:VIMConnectionNameComments];
 
-    if (self.canViewComments)
-    {
-        if ([commentsConnection.total respondsToSelector:@selector(intValue)])
-        {
-            return commentsConnection.total.intValue;
-        }
-        else
-        {
-            NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
+    if (self.canViewComments == NO)
+        return 0;
 
-            return 0;
-        }
+    if ([commentsConnection.total respondsToSelector:@selector(intValue)])
+    {
+        return commentsConnection.total.intValue;
     }
     else
     {
+        NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
+
         return 0;
     }
 }
--- a/Sources/Shared/Models/VIMVideo.m
+++ b/Sources/Shared/Models/VIMVideo.m
@@ -541,8 +541,24 @@ NSString *VIMContentRating_Safe = @"safe";
 - (NSInteger)commentsCount
 {
     VIMConnection *commentsConnection = [self connectionWithName:VIMConnectionNameComments];
-    
-    return (self.canViewComments ? commentsConnection.total.intValue : 0);
+
+    if (self.canViewComments)
+    {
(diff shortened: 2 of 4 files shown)


Jumhyn:
@van Can you open this as a PR against https://github.vimeows.com/MobileApps/VimeoNetworking as well?

Sources/Shared/Models/VIMVideo.m
```diff
@@ -541,8 +541,24 @@ - (NSInteger)likesCount
...
+
+    if (self.canViewComments)
+    {
+        if ([commentsConnection.total respondsToSelector:@selector(intValue)])
+        {
+            return commentsConnection.total.intValue;
+        }
+        else
+        {
+            NSAssert(false, @"Error: Detected instance where `NSNumber` is not `NSNumber`.");
```
nicolelehrer: nit pick - "...where `commentsConnection.total` is unexpectedly not an `NSNumber`"

Sources/Shared/Models/VIMVideo.m
```diff
@@ -541,8 +541,24 @@ - (NSInteger)likesCount
 - (NSInteger)commentsCount
 {
     VIMConnection *commentsConnection = [self connectionWithName:VIMConnectionNameComments];
-    
-    return (self.canViewComments ? commentsConnection.total.intValue : 0);
+
+    if (self.canViewComments)
```
rpassis: how about getting rid of some of this nesting by returning early?

```
if (self.canViewComments == NO) { return 0; }
...
```

nicolelehrer (approved):
Code looks good - worth adding a unit test? otherwise can test when we link with client

rpassis (changes requested):
Hey I think this fixes the crash but I wonder if the comments count method is actually doing what it is supposed to do here.

Based on the API docs https://developer.vimeo.com/api/reference/videos#get_comments it may be that the connection we are looking for is not called `comments` but `replies`.

Worth checking with the API team?

supperspidey:
I'm not sure how I'll be able to unit test this change because that'll require us to somehow confuse the object mapping code so that it can map the `total` property to a type different from `NSNumber`. I'll change the assertion message though. Thanks!

supperspidey:
@Jumhyn once the release branch is merged, @rpassis will cherry pick this change to that repo per our conversation.

Jumhyn:
Another thought--since we don't really know what's going on here we might be better served by catching this at deserialization-time by adding a `-didFinishMapping` implementation to `VIMConnection`:
```
- (void)didFinishMapping {
    if ([self.total isKindOfClass: [NSNumber class]] == NO) {
        NSAssert(false, ...);
        self.total = @0;
    }
}
```

cameo-github:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 231 tests, with 0 failures (0 unexpected) in 1.494 (1.970) seconds</td>
    </tr>
  </tbody>
</table>

See build details on [CircleCI](https://circleci.com/gh/vimeo/VimeoNetworking/1669)

</p>

nicolelehrer:
@Jumhyn I like that but not sure if `didFinishMapping` gets called by `VIMConnection`? See `(void)parseConnections` where we have to init with dictionary, I think this unfortunately bypasses the flow

rpassis:
> I'm not sure how I'll be able to unit test this change because that'll require us to somehow confuse the object mapping code so that it can map the total property to a type different from NSNumber. I'll change the assertion message though. Thanks!

This made me rethink some of what we are doing here. 

Based on your observation above, do you think that perhaps what needs fixing is not the `commentsCount` method on `VIMVideo` but the actually parsing logic on `VIMConnection`?

Jumhyn:
I believe it should—line 286 of VIMVideo.m has this:
```
if([connection respondsToSelector:@selector(didFinishMapping)])
    [connection didFinishMapping];
```

There's a similar call on VIMUser, as well.

supperspidey:
@Jumhyn I like your approach too. That way, we can catch any faux `total` property much early on. Plus, I can also remove the additional fix in the client app. It's much cleaner.

nicolelehrer:
Ah I see it @Jumhyn. We should def check it gets called @supperspidey 
I like getting to the root of this as y'all suggest but playing devil's advocate if this is just a patch, is we lose context of where this is failing since we're not properly logging these errors yet. @supperspidey might want to just check with API concurrent to making this change to check our assumptions.

nicolelehrer:
But not opposed to catching `total` errors for all connections 👍

supperspidey:
> Hey I think this fixes the crash but I wonder if the comments count method is actually doing what it is supposed to do here.
> 
> Based on the API docs https://developer.vimeo.com/api/reference/videos#get_comments it may be that the connection we are looking for is not called `comments` but `replies`.
> 
> Worth checking with the API team?

I've just checked the API doc. The comments are still being referred to as `metadata.connection.comments`, just like how our code is doing:

https://developer.vimeo.com/api/reference/responses/video

supperspidey:
Thanks guys for your feedback! Ready when you are.

Sources/Shared/Models/VIMConnection.m
```diff
@@ -83,4 +83,12 @@ - (BOOL)canPost
     return (self.options && [self.options containsObject:@"POST"]);
 }
 
+- (void)didFinishMapping
```
Jumhyn, line 86: Now that `VIMConnection` has a `-didFinishMapping` method, we'll want to make sure that we call up to `super` from `VIMNotificationsConnection`'s implementation (and any other subclass implementations, but I think that's the only one).

Sources/Shared/Models/VIMConnection.m
```diff
@@ -83,4 +83,12 @@ - (BOOL)canPost
...
+{
+    if (self.total != nil && [self.total isKindOfClass: [NSNumber class]] == NO) {
+        NSAssert(false, @"Error: Detected instance where `commentsConnection.total` is unexpectedly not an `NSNumber`.");
```
nicolelehrer: `connection.total` instead of `commentsConnection.total`

Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -1014,7 +1013,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 1100;
-				LastUpgradeCheck = 1130;
+				LastUpgradeCheck = 1100;
```
nicolelehrer: Are these changes from a pod install? any concern we went down?

supperspidey: Good question. I think we don't need to commit the Pods project file in order to make the unit tests run, as I initially thought. Let me see if I can revert this change.

supperspidey: The file has been reverted! Thanks Nicole!

rpassis (approved):

nicolelehrer (approved):

supperspidey:
Thanks guys!

PR #416, opened 2020-04-06
Check property type before setting new value

roman-khasanov:
#### Ticket

[990](https://vimean.atlassian.net/browse/VCI-990)

#### Issue Summary

Crash appears in VIMObjectMapper when a property is updated with an object of inappropriate type.

if (keyFound == NO) {
    [newObject setValue:jsonValue forKey:objectKey];
}
-[__NSDictionary0 _fastCStringContents:]: unrecognized selector sent to instance 

#### Implementation Summary

Added a check of the property type before setting new value.

#### How to Test

We haven't managed to reproduce the crash so I think we can just emulate wrong JSON.

### DIFF
--- a/Sources/Shared/Models/VIMObjectMapper.m
+++ b/Sources/Shared/Models/VIMObjectMapper.m
@@ -26,6 +26,16 @@
 
 #import "VIMObjectMapper.h"
 #import "VIMMappable.h"
+#import <objc/runtime.h>
+
+enum VimeoPropertyType {
+    kVimeoPropertyTypeNSString,
+    kVimeoPropertyTypeNSDictionary
+};
+
+typedef enum VimeoPropertyType VimeoPropertyType;
+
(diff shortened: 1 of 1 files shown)


gennadiiivanov (changes requested):
@roman-khasanov what if class is not `NSString ` but `NSAttributedString` or any other class?

Jumhyn:
@gennadiiivanov This looks like it's addressing the same issue that @supperspidey has an open PR for [here](https://github.vimeows.com/MobileApps/Models/pull/33). @roman-khasanov give a look over there to see how Van approached it.

roman-khasanov:
> @gennadiiivanov This looks like it's addressing the same issue that @supperspidey has an open PR for [here](https://github.vimeows.com/MobileApps/Models/pull/33). @roman-khasanov give a look over there to see how Van approached it.

👍

